PR #121, opened 2016-10-25
Added isFloat

aarongerig:
Adds the ability to receive the number type provided easily. Like this you can adjust your number slider's step amount, which you are using as a filter.

Sometimes you receive decimal numbers from your Pimcore object and sometimes you receive integers. Therefore you need the ability to adjust your step amount of your number slider accordingly. See code below:

```php
<input title="<?= $this->translate($this->label); ?>"
                  type="text"
                  name="<?= $this->fieldname; ?>"
                  class="range-slider span2"
                  value=""
                  id="<?= $this->fieldname; ?>Slider"
                  data-slider-id="<?= $this->fieldname; ?>Slider"
                  data-slider-min="<?= $this->minValue; ?>"
                  data-slider-max="<?= $this->maxValue; ?>"
                  data-slider-step="<?= $this->isFloat ? '0.1' : '1'; ?>"
                  data-slider-value="[<?= $this->currentValueMin; ?>, <?= $this->currentValueMax; ?>]">
```

### DIFF
--- a/lib/CoreShop/Model/Product/Filter/Condition/Range.php
+++ b/lib/CoreShop/Model/Product/Filter/Condition/Range.php
@@ -85,13 +85,20 @@ class Range extends AbstractCondition
         $rawValues = $list->getGroupByValues($this->getField(), true);
         $script = $this->getViewScript($filter, $list, $currentFilter);
 
+        $product = [];
+        foreach (['value'] as $value) {
+            $product[$value] = array_product(array_column($rawValues, $value));
+        }
+
         $minValue = count($rawValues) > 0 ? $rawValues[0]['value'] : 0;
         $maxValue = count($rawValues) > 0 ? $rawValues[count($rawValues)-1]['value'] : 0;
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
Well, but maybe if the number is a float, you want a step count of 1. Maybe we should extend the Range Filter Settings to add a field for step count?

aarongerig:
Sure, that might be the case, but you don't have to use the `isFloat` conditional statement in your view if not needed... The change I made only represents a extension to the method `render`.

If adding a additional settings field for step count or not, is only a matter of opinion, whether you want to make the change programmatically or possible to be edited by the user.

dpfaffenbauer:
I just think that a step count of 0.1 or 1 doesn't fit every use case. So making this as a setting would be more generic.

aarongerig:
You're right, it's way more flexible that way – I just tried to solve it programmatically. But if you can provide another settings field, that'd be great!

dpfaffenbauer:
see 8ecf24e7fa8aa51b3a32fbb241cc5961955b6adc

PR #182, opened 2017-08-23
[Migrations] take advantage of Pimcore Migrations

dpfaffenbauer:


Yivan:
Just why using Twig_Filter as it seems deprecated, please see:
https://twig.symfony.com/api/1.x/Twig_Filter.html
https://twig.symfony.com/api/1.x/Twig_SimpleFilter.html

dpfaffenbauer:
Have a look there: http://twig.readthedocs.io/en/1.x/deprecated.html

PS: we are using Twig 2.* it's a bit confusing. First they deprecate it, then they don't :/

Yivan:
You are right, i missed the twig version ; ) Yes it is a bit confusing.
Thanks!

PR #238, opened 2017-12-18
CartContext instead of CartManager

dpfaffenbauer:
This PR changes the behaviour of how CoreShop finds the current active cart. The CartManager now is only used for persisting the cart and cart-items.


src/CoreShop/Bundle/OrderBundle/Resources/config/services/context.yml
```diff
@@ -0,0 +1,42 @@
+services:
+  coreshop.context.cart.new:
```
solverat: maybe put them in right order so its easier to get the priority list.

PR #261, opened 2018-01-04
[FrontendBundle] make reset password url absolute

ktallafus:
The default parameter for generateUrl is UrlGeneratorInterface::ABSOLUTE_PATH, not ABSOLUTE_URL

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Controller/RegisterController.php
+++ b/src/CoreShop/Bundle/FrontendBundle/Controller/RegisterController.php
@@ -21,6 +21,7 @@ use CoreShop\Component\Address\Model\AddressInterface;
 use CoreShop\Component\Customer\Model\CustomerInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 
 class RegisterController extends FrontendController
 {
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
doesn't Pimcore Mail already do that?

ktallafus:
no, when I tested it, it didn't do that

dpfaffenbauer:
have you set the main-domain in pimcore settings?

ktallafus:
It wasn´t, but when I set it, the link is still not absolute

dpfaffenbauer:
alright, I'll merge it, but its something that pimcore somehow already does as well.

dpfaffenbauer:
thanks @ktallafus

ktallafus:
I also asked solverat about this, but he said "nicht dass ich wüsste nein" ;)
thanks!

dpfaffenbauer:
well, so he doesn't know about it: https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Mail.php#L729

ktallafus:
Yes, you´re right. I debugged it. But it seems like the hostUrl has to be given as option to the Mail constructor:
https://github.com/pimcore/pimcore/blob/master/pimcore/lib/Pimcore/Mail.php#L198

This doesn´t happen here:
https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Component/Notification/Rule/Action/MailActionProcessor.php#L47

So... it seems my commit was wrong :(

dpfaffenbauer:
Not completely right either: Pimcore has a "fallback" for that: https://github.com/pimcore/pimcore/blob/856e9bf14c19d82dc2bd48b762e1f9435b7bef14/pimcore/lib/Pimcore/Helper/Mail.php#L229

ktallafus:
Ok, I got it. The problem was the link itself. I made the link in wysiwyg and it prepended the http:// protocoll. So the href ended up as http:///customer/password-reset (notice the three slashes). So Pimcore couldn´t identify it as link. If i select the protocoll to "Other", it works.

So, are you going to revert this?

dpfaffenbauer:
thanks for looking deeper into that, I now reverted the PR. Thanks anyway :)

PR #274, opened 2018-01-09
[FrontendBundle] Registration can redirect back to checkout

ktallafus:
When registration is called from checkout process, it should redirect back when done.


dpfaffenbauer:
many thanks

PR #275, opened 2018-01-09
[FrontendBundle] Customer Address can redirect back to checkout

ktallafus:
When new address is added from checkout process, it should redirect back when done.


dpfaffenbauer:
thanks

PR #300, opened 2018-01-31
Listing improvment

solverat:
- new option for store config (category list view): "List: Include Subcategories". Allows to include subcategories if no filter has been defined
- implemented new `getProducts` method to product repository
- fix `findLatestByStore` method
- use `addConditionParam` instead of `setCondition` in `findBy()` method to allow multiple conditions for listing


dpfaffenbauer:
cheers

PR #303, opened 2018-02-07
Refactor filter

dpfaffenbauer:
- Adds nested Filter
 - fix for nested conditions all around CoreShop (prices, shipping, etc)
 - remove field "field" as required from filter
 - cleanups


solverat:
- "field" stores nested type as "Array". 
- hide label in extjs, if it's in a nested route?

dpfaffenbauer:
field: is deprecated anyway, but shouldn't be persisted anymore, will look into it.
label: well, thats a UI thing and can't be solved a generic way. guess we need to solve it depending on the project.

PR #315, opened 2018-02-13
Fixed search condition filtering only products set as active

GALCF:
Fixes search controller returning all active products regardless of the search because active is set in an "OR" condition.


dpfaffenbauer:
thank you :)

PR #332, opened 2018-02-22
Translate tax display

Cruiser13:


dpfaffenbauer:
thats not the right way to do it, this could cause issues when using different languages. a better approach would be something like:

Key: ```coreshop_product_tax_inc```
Value: ```incl. %s VAT```

So, it would be written like this:

```
'coreshop_product_tax_inc|trans|format(product|coreshop_product_tax_rate) 
```

dpfaffenbauer:
Can you add the translation definition as well? https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Bundle/FrontendBundle/Resources/install/pimcore/translations.yml

dpfaffenbauer:
ping @Cruiser13 can you please add the missing translations here?

https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Bundle/FrontendBundle/Resources/install/pimcore/translations.yml

Cruiser13:
Bulk translations added here
https://github.com/coreshop/CoreShop/pull/358

dpfaffenbauer:
thank you :) 🍻

Issue #335, opened 2018-02-24
Clarify Country Context Resolver 


solverat:
### Available Country Resolver

Resolver with the highest priority will dispatch the current country:

| Name | Priority | Comment
|-------|--------|--------|
| Store Resolver | 100 | _Priority should switch with Country Resolver?_ |
| Country Resolver (Geo IP) | 10 | _Priority should switch with Store Resolver?_ |
| Cart Resolver | 1 |  |

### Country Resolver in Conditions
There are two Price Rule Conditions: `country` and `zone`. Both conditions also uses the current resolved country **not** the current user location. 

### Product Tax Calculator
Product Tax Calculator is also bounded to the current resolved country.

### Todo
- Really switch Priority?
- Build Solid Scenarios 
- Stores before Country?
- Tax Rules from Store or from Country Resolver?
- If Store only allows one Country, Currency and Tax-Rule: How to serve the right context to customer if the priority gets switched (which would mean the current store gets ignored)?
- bind country context to valid countries from store

PR #338, opened 2018-02-26
Refactor Country Context

dpfaffenbauer:
Valid Country now resolves like this:

 - Cart Shipping Address
 - GEO IP

If one of the above is not configured inside the store, the store-base country is the valid one.


solverat:
this PR resolves #335.

Issue #178, opened 2017-07-26
[CoreShop2] Front-End site exemple template in PHP.


Yivan:
Hello,

Is it planned to have the frontend site exemple in PHP, or it will be only Twig ?
Having it in PHP would give more insight to make a frontend using PHP template, as documentation is light on the subject (for CoreShop 1, the website exemple is very helpfull to design the front end by giving concrete PHP exemple).
Having it in PHP won't refrain people who don't want to use twig to not using CoreShop.
I think having one in PHP and one in Twig would be really great.

To be honest our agency don't use Twig for now, prefering to work with PHP for template. Having only Twig frontend could really be a no-go for using CoreShop, but we love CoreShop : ) So if on the stable release Twig and PHP front end could be available it will be really a good thing!

Thanks.

dpfaffenbauer:
Hi Yivan,

currently it is not planned, but I'm happy to accept PR's on this topic. Since CoreShop isn't specifically limited to Twig, its possible to copy the FrontendBundle and manually convert all Templates to PHP Engine. But since I have no time currently, I can't provide PHP Engine templates. It's also kinda difficult for me to provide two FrontendBundles with two different Engines, that would mean twice the work for me then as well.

And, don't forget that the FrontendBundle is meant as demo/example and not really as kind of a "default template", if you know what I mean. It should only help you getting along with CoreShop. So I can't really see the major benefit in having TWIG and PHP.

I'll leave this issue open anyway for people who are interested in PHP Engine as well.

Yivan:
Thanks for you answer @dpfaffenbauer 

It is a very sad news for us. Our team have to learn the new Pimcore5, Symfony3 and CoreShop2 for the migration. It makes a lot already to learn for them! Adding Twig on the stack (we don't use it for now on our projects) will be problematic, we prefer to stick to PHP template for now (maybe Twig later).
It mean we will loose one of the precious advantage of Coreshop : having a ready front-end.

As you say it, front-end is here just for example, but it is a really valuable addition for starting project and undestand the way coreshop works. We see the front-end bundle as a boilerplate template to start project.
In providing only twig, we won't be able to use CoreShop2 now i think : (( Maybe later after the team is more expert with Pimcore5/Symfony3.

I undestand perfectly it is lot of work to maintain 2 versions, but I really hope you could find a way to provide both one day. Having both will not prevent agency/people like us not using twig to not using CoreShop2 (or being reluctant because it need to re-make the front-end if not using twig).

[...]

PR #362, opened 2018-03-08
[FrontendBundle] Template Configurator 

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #178

implement TemplateConfigurator to allow views to be loaded from different locations and in different engines. This allows CoreShop to be used with the PHP Engine as well. You can now configure following things:

```
core_shop_frontend:
    view_bundle: 'App'
    view_suffix: 'php'
```

This changes the template location path to @App and the suffix to php, so for the Cart Summary that would be: '@App/Cart/summary.html.php'

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Resources/config/services.yml
+++ b/src/CoreShop/Bundle/FrontendBundle/Resources/config/services.yml
@@ -2,79 +2,62 @@ imports:
     - { resource: "services/wishlist.yml" }
 
 services:
-    coreshop.frontend.controller.index:
-        class: '%coreshop.frontend.controller.index%'
+    coreshop.frontend.controller.abstract:
+        class: CoreShop\Bundle\FrontendBundle\Controller\FrontendController
+        abstract: true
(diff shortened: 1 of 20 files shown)


src/CoreShop/Bundle/FrontendBundle/Resources/config/services.yml
```diff
@@ -6,63 +6,75 @@ services:
         class: '%coreshop.frontend.controller.index%'
         calls:
             - [setContainer, ['@service_container']]
+            - [setTemplateConfigurator, ['@coreshop.frontend.template_configurator']]
```
solverat: Use the `parent` attribute to define the parent controller so we don't need to add all those calls?

dpfaffenbauer: done

Issue #377, opened 2018-03-23
Implement Salutation Field


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

- add `salutation` to customer
- add `salutation` to address
- allow to define salutation prefixes in every country

PR #378, opened 2018-03-24
Salutation Fields

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #377 

This PR adds a salutation field (new form type `SalutationChoiceType`) to the Customer and Address-Form. All salutation prefixes are stored in every country entity. It also updates the address format by adding the salutation per default. Since we're handling a choice type here, the stored value also gets translated in the `AddressFormatter` class.

## Open Questions / Statements
- If someone wants to add a additional salutation prefix for every country: Is it reasonable to change the choice list via form extension for `SalutationChoiceType`?
- The prefix list is bounded to the country context. If there is no country there is also no salutation prefix
- Since the salutation prefixes may change after a user changes the country via the country choice type (only in address), we need to implement some ajax logic (and also a public accessible method to get salutation prefixes of a selected country)


dpfaffenbauer:
cheers

Issue #370, opened 2018-03-22
Calculate Shipping Costs in Cart


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

## Overview / Additionals
- Provide a simple Form (ZIP + Country) to calculate shipping costs in cart.
- Update Docs 

## Calculation Workflow, sorted by Priority
1. **Prio 300**: If Cart has shipping address, use shipping address to calculate
2. **Prio 200**: If Customer/Guest + default address is available, use default address to calculate
3. **Prio 100**: If Cart is not empty, provide a simple form (zip, country) to calculate shipping costs

solverat:
great work 🎉

PR #379, opened 2018-03-24
Carrier Calculation in Cart

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #370 

This PR adds some UX improvements in cart to get available carriers by adding a new form (`ShippingCalculatorType`).

See it in action here: http://recordit.co/hskjM1mGKE

## Statements
- If a cart already has some calculated shipment values (happens if customer already has chosen a shipment address and a carrier in the carrier checkout step), the form will **no longer** show up. Otherwise it's quite confusing if there are shipment prices in the cart overview and also a optional carrier selection form

## Questions
- @dpfaffenbauer as already discussed, there is still some work to do: If a user checks the available carriers within this new form, he should be able to select the available carrier. add the carrier to the cart should be enough? are there other ways to provide the cheapest carrier on the fly? Please fill free to add the necessary adoptions to this PR

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Migrations/Version20180323141834.php
+++ b/src/CoreShop/Bundle/CoreBundle/Migrations/Version20180323141834.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace CoreShop\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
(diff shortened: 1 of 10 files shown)


dpfaffenbauer:
cheers

Issue #380, opened 2018-03-24
Change Response Type based on Request Type


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Allow different response types depending on request type (rest bundle, serializer). Allow to still determinate request type (sometimes we need to get html as a response in ajax requests).

## Use Cases

- update cart via ajax
- update cart widget via ajax
- change salutation prefixes depending on country
- change address state based on country
- cart carrier calculation (currently a html response which replaces the given form)

PR #384, opened 2018-03-26
[WIP] implement view agnostic rendering

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #380 

This PR allows to use view-agnostic rendering. That means, based on the accept http header, you either get a JSON Response or HTML Response. This PR is not finished, as the checkout for example is not working yet. Also: It still not defined on how to activate this or not, cause in most cases, you don't need to have this enabled. My idea would be to create a route-configuration, but that means moving away from static routes to symfony routes.


solverat:
> "My idea would be to create a route-configuration, but that means moving away from static routes to symfony routes."

There is nothing bad about symfony routes: <3

PR #392, opened 2018-04-05
Proposal: Product display restriction by store

Cruiser13:
Disallow product rendering if a store is selected which is not allowed in the product.

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Controller/ProductController.php
+++ b/src/CoreShop/Bundle/FrontendBundle/Controller/ProductController.php
@@ -48,7 +48,7 @@ class ProductController extends FrontendController
             throw new NotFoundHttpException('product not found');
         }
         
-        if(!in_array($this->get('coreshop.context.store')->getStore()->getId(),$product->getStores())){
+        if(!in_array($this->get('coreshop.context.store')->getStore()->getId(), $product->getStores())){
             throw new NotFoundHttpException('product not found');
         }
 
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/FrontendBundle/Controller/ProductController.php
```diff
@@ -47,6 +47,10 @@ public function detailAction(Request $request)
         if (!$product->isPublished() || $product->getActive() !== true) {
             throw new NotFoundHttpException('product not found');
         }
+        
+        if(!in_array($this->get('coreshop.context.store')->getStore()->getId(),$product->getStores())){
```
dpfaffenbauer: looks good, except for your code formatting. Can you please fix that? (What I mean is the missing whitespace after the comma)

Cruiser13:
I don't know if there are other places where stores are not properly checked for (search maybe?) too.

src/CoreShop/Bundle/FrontendBundle/Controller/ProductController.php
```diff
@@ -47,6 +47,10 @@ public function detailAction(Request $request)
         if (!$product->isPublished() || $product->getActive() !== true) {
             throw new NotFoundHttpException('product not found');
         }
+        
+        if(!in_array($this->get('coreshop.context.store')->getStore()->getId(), $product->getStores())){
```
dpfaffenbauer: don't want to bother you, but a whitespace after the if is missing as well :D

Issue #398, opened 2018-04-06
Backend Order Creation does not save language


Cruiser13:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Steps to reproduce:
Create a new order and set a different language, german for example:
![order-german](https://user-images.githubusercontent.com/2771909/38425582-8fa8d4d4-39b4-11e8-9781-7fc329758364.png)
It'll not be saved and the default language will be assigned to the order:
![order-english](https://user-images.githubusercontent.com/2771909/38425597-9db1f466-39b4-11e8-83be-a89fb0184a72.png)

This leads to invoices in the wrong language and further issues.

I believe that language is not properly assigned here: https://github.com/coreshop/CoreShop/blob/3b63bec32bad078aaab24d61f04f45ba0baa7d71/src/CoreShop/Bundle/OrderBundle/Controller/AbstractSaleCreationController.php#L303
It's being saved before as far as I could debug this: https://github.com/coreshop/CoreShop/blob/3b63bec32bad078aaab24d61f04f45ba0baa7d71/src/CoreShop/Bundle/OrderBundle/Controller/AbstractSaleCreationController.php#L234

dpfaffenbauer:
this commit should fix it: https://github.com/coreshop/CoreShop/commit/d825a2d03f77632da38be14d905487fe24bafdf5

What it changes:
 - Use Pimcore Locale Fallback if not fallback is found by using `findLocale`instead of `getLocale` on Pimcores locale Service
 - Don't use the LocaleContext inside the SaleTransformer, store the locale inside the cart and use it for order-creation.

dpfaffenbauer:
this is fixed now

PR #399, opened 2018-04-08
Refactor SaleTransformer to not use Contexts anymore

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/kinda
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #398


src/CoreShop/Component/Order/Model/Cart.php
```diff
@@ -226,4 +226,20 @@ public function setPaymentProvider($paymentProvider)
     {
         throw new ImplementedByPimcoreException(__CLASS__, __METHOD__);
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLocaleCode()
```
solverat: we do have three different locale getters now: 

``` 
cart: getLocaleCode
order: getOrderLanguage
user: getLocale
```
mark them as deprecated and use consistent getters like `getLocale()` could be a good idea. But of course we also need to take care about a good migration process then.

src/CoreShop/Component/Order/Transformer/AbstractCartToSaleTransformer.php
```diff
@@ -158,7 +149,7 @@ public function transformSale(ProposalInterface $cart, ProposalInterface $sale,
         $sale->setPublished(true);
         $sale->setParent($orderFolder);
         $sale->setCustomer($cart->getCustomer());
-        $sale->setSaleLanguage($this->localeContext->getLocaleCode());
+        $sale->setSaleLanguage($cart->getLocaleCode());
```
solverat: existing carts never gets the locale applied, so this could lead to some troubles here. we should add the locale to existing carts in the migration process.

src/CoreShop/Component/Order/Model/Cart.php
dpfaffenbauer: done here #403

src/CoreShop/Component/Order/Transformer/AbstractCartToSaleTransformer.php
dpfaffenbauer: migration added https://github.com/coreshop/CoreShop/pull/399/files#diff-cfdd1e300775b2ffa2224f3bc9b0294dR23

Issue #414, opened 2018-04-15
Store Product with Price 0.00 not get stored


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Products with Price `0.00` get skipped during the save process (db value = `null`). Since a product also could be a container product, they should get stored in db regardless its price.

dpfaffenbauer:
doesn't that just happen on migrations?

I don't see any condition on saving them: https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Bundle/CoreBundle/CoreExtension/StorePrice.php#L248

PR #417, opened 2018-04-17
[CoreBundle] reset original value of store prices on save

dpfaffenbauer:
reset original value of store prices on save -> these prohibited to save values after they have been changed once.

| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #414


dpfaffenbauer:
@solverat Not a 100% sure if this fixes your issue though

PR #419, opened 2018-04-18
Attach customer and order object to event listener

albertmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no


dpfaffenbauer:
not necessary, the MailActionProcessor does that here: https://github.com/coreshop/CoreShop/blob/6772582c15f9ec604a4302889a1b7b4cc1a60e7d/src/CoreShop/Component/Notification/Rule/Action/MailActionProcessor.php#L48

PR #423, opened 2018-04-24
Put the categoryId between commas

albertmueller:
Because in current method, will be search in the whole string

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


albertmueller:
It would be also possible to change the logic from "LIKE" to a "IN" condition.

dpfaffenbauer:
From LIKE to IN is not possible, since it is only a string.

Issue #413, opened 2018-04-12
Nice Urls Macro


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

The default coreshop routing creates ugly urls like: `shop/fr/Monogram%20Canvas%20Montaigne%20BB~p62`.

## Suggestion
We should implement a macro file which allows devs to create nice looking urls without changing all the templates. Affected routes:

- `coreshop_product_detail`
- `coreshop_category_list`

So instead of 

```twig
{{ path('coreshop_category_list', {name: category.getName, category: category.getId}) }}
```

we should use

```twig
{% import '@CoreShopFrontend/Common/Macro/routing.html.twig' as routing %}
{{ routing.nice_path('coreshop_category_list', {name: category.getName, category: category.getId}) }}
```

By default the macro will return the same url to prevent BC breaks.

## Open Question
What about routes which get generated via api (without template engine), they also should get transformed the same way.

dpfaffenbauer:
Only viable solution would be a template helper that generates routes + a separate service. What do you think about that?

dpfaffenbauer:
btw, you can already do that anyway. Why create a separate service for that when this is already symfony standard?

solverat:
After some research. Basically we should use the [pimcore link generator](https://pimcore.com/docs/5.x/Development_Documentation/Objects/Object_Classes/Class_Settings/Link_Generator.html).

## Service
First, we need a global CoreShop service called `LinkGenerator`. This one can be injected and used to create urls in php context (links in mails for example):

```php
<?php

class LinkGenerator
{
    public function getLink(Concrete $entity, string $name, array $params)
    {
        $params['routeName'] = $name;
        $params['entity'] = $entity;

        if ($linkGenerator = $entity->getClass()->getLinkGenerator()) {
            $_params = array_merge($params, [
                'document'     => $this->documentResolver->getDocument();
                'context'      => $this
            ]);

            return $linkGenerator->generate($entity, $_params);
        }

        // throw exception if no link generator has been found for given class.
    }
}
```

## Twig
Create a Twig Extension (which calls `getLink()` from the service above) to generate links in templates:

```twig
{{ coreshop_link(product, 'coreshop_product_detail') }}
```

## Link Generator
Then CoreShop should provide a link generator service for each class (`product`, `category` should be enough?)  Add the services to the link generator in pimcore class configuration.
For example the product link generator

```php
[...]

PR #452, opened 2018-05-23
[UrlGenerator] create a wrapper url generator

dpfaffenbauer:
in order to have the same urls in code than in templates and still use Pimcore Link Generator. This also makes it easier for devs to change certain url layouts without touching all templates

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #413


dpfaffenbauer:
@solverat I am not a 100% sure about the slug thing here. This introduces sort of a BC break cause it changes URL's and that affects SEO...

src/CoreShop/Component/Pimcore/Templating/Helper/LinkGeneratorHelper.php
```diff
@@ -0,0 +1,116 @@
...
+
+        return $this->linkGenerator->generate($object, $routeName, $params, $relative ? UrlGeneratorInterface::NETWORK_PATH : UrlGeneratorInterface::ABSOLUTE_URL);
+    }
+
+    /**
+     * @param $arguments
+     * @return array
+     */
+    protected function prepareParameters($arguments)
+    {
```
solverat: Extending the default symfony helper class so we don't neeed to add this method?

dpfaffenbauer: this method allows you to do following thing:

```twig
coreshop_path($object, 'route', {}, false)
//and
coreshop_path('route', {}, false)
//and
coreshop_path($object)
//and
coreshop_path('route_name')
```

So, it checks what kind of arguments come and therefore we don't need to call the function without a `null`parameter for non object-routes.

src/CoreShop/Bundle/CoreBundle/Migrations/Version20180523085857.php
```diff
@@ -0,0 +1,46 @@
...
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class Version20180523085857 extends AbstractPimcoreMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
```
solverat: That's a bc break right? What about just adding a upgrade note? As long no link generator has been defined the default pathgenerator gets triggered => no bc break

dpfaffenbauer: Well, do you mean cause of the slug? Otherwise: No. The DataObjectLinkGenerator does the same as the old version did except for the slug -> which IMO is a BC break. We could remove the migration, would be fine with me.

solverat: just checked it. the migration is fine since you need to change the twig extension from `path` to `coreshop_path` anyway. If you keep the `path` the url isn't changing so there is no bc break after all. 👍

dpfaffenbauer: Good.

dpfaffenbauer: BUT: What if you didn't override the template?

solverat: you're out of luck.

PR #474, opened 2018-06-15
SilberMa: Change active, digitalProduct, isTracked from checkbox to b…

SilberMa:
…ooleanselect to solve variant and inheritance issues

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Migrations/Version20180615093141.php
+++ b/src/CoreShop/Bundle/CoreBundle/Migrations/Version20180615093141.php
@@ -12,34 +12,34 @@ class Version20180615093141 extends AbstractPimcoreMigration implements Containe
 {
     use ContainerAwareTrait;
 
-    private const CORE_SHOP_PRODUCT_FIELDS = [
-        [
-            'fieldName' => 'active',
-            'yesValue'  => 'active',
-            'noValue'   => 'inactive',
-        ],
-        [
-            'fieldName' => 'digitalProduct',
-            'yesValue'  => 'isDigitalProduct',
-            'noValue'   => 'noDigitalProduct',
-        ],
-        [
-            'fieldName' => 'isTracked',
-            'yesValue'  => 'isTracked',
-            'noValue'   => 'notTracked',
-        ],
-    ];
-
     /**
      * @param Schema $schema
      */
     public function up(Schema $schema)
     {
+        $coreShopProductFields = [
+            [
+                'fieldName' => 'active',
+                'yesValue'  => 'active',
+                'noValue'   => 'inactive',
+            ],
+            [
+                'fieldName' => 'digitalProduct',
+                'yesValue'  => 'isDigitalProduct',
+                'noValue'   => 'noDigitalProduct',
+            ],
+            [
+                'fieldName' => 'isTracked',
+                'yesValue'  => 'isTracked',
+                'noValue'   => 'notTracked',
+            ],
+        ];
+
         $coreShopProductClass = $this->container->getParameter('coreshop.model.product.pimcore_class_name');
 
         $classUpdater = new ClassUpdate($coreShopProductClass);
 
-        foreach (self::CORE_SHOP_PRODUCT_FIELDS as $updateField) {
+        foreach ($coreShopProductFields as $updateField) {
             if ($classUpdater->hasField($updateField['fieldName'])) {
                 $classUpdater->replaceFieldProperties(
                     $updateField['fieldName'],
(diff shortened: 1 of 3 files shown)


src/CoreShop/Bundle/CoreBundle/Migrations/Version20180615093141.php
```diff
@@ -0,0 +1,87 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+
+class Version20180615093141 extends AbstractPimcoreMigration implements ContainerAwareInterface
+{
+    use ContainerAwareTrait;
+
+    private const CORE_SHOP_PRODUCT_FIELDS = [
```
dpfaffenbauer: `private const`  is not supported in PHP 7.0, anyway, a const is not even necessary here, can't you just create the variable inline and use it then?

src/CoreShop/Bundle/CoreBundle/Resources/install/pimcore/classes/CoreShopProductBundle/CoreShopProduct.json
```diff
@@ -210,8 +210,36 @@
...
                                     },
                                     {
-                                        "fieldtype": "checkbox",
-                                        "defaultValue": 0,
+                                        "fieldtype": "booleanSelect",
+                                        "yesLabel": "isDigitalProduct",
+                                        "noLabel": "noDigitalProduct",
+                                        "emptyLabel": null,
+                                        "options": [
+                                            {
```
dpfaffenbauer: not valid json here, should be like this:

```json
"options": [
    {
        "key": "empty",
        "value": 0
    },
    {
        "key": "isTracked",
        "value": 1
    },
    {
        "key": "notTracked",
        "value": -1
    }
]
```

src/CoreShop/Bundle/CoreBundle/Resources/install/pimcore/classes/CoreShopProductBundle/CoreShopProduct.json
```diff
@@ -645,8 +673,36 @@
                                 "visibleSearch": false
                             },
                             {
-                                "fieldtype": "checkbox",
-                                "defaultValue": 0,
+                                "fieldtype": "booleanSelect",
+                                "yesLabel": "isTracked",
+                                "noLabel": "notTracked",
+                                "emptyLabel": null,
+                                "options": [
```
dpfaffenbauer: not valid json here, should be like this:

```json
"options": [
    {
        "key": "empty",
        "value": 0
    },
    {
        "key": "isTracked",
        "value": 1
    },
    {
        "key": "notTracked",
        "value": -1
    }
]
```

src/CoreShop/Bundle/ProductBundle/Resources/install/pimcore/classes/CoreShopProduct.json
```diff
@@ -188,8 +188,36 @@
                                         "visibleSearch": true
                                     },
                                     {
-                                        "fieldtype": "checkbox",
-                                        "defaultValue": 0,
+                                        "fieldtype": "booleanSelect",
+                                        "yesLabel": "active",
+                                        "noLabel": "inactive",
+                                        "emptyLabel": null,
+                                        "options": [
```
dpfaffenbauer: not valid json here, should be like this:

```json
"options": [
    {
        "key": "empty",
        "value": 0
    },
    {
        "key": "isTracked",
        "value": 1
    },
    {
        "key": "notTracked",
        "value": -1
    }
]
```

dpfaffenbauer:
These works fine now. The only issue I see now is, and we had that before as well, that Pimcores BooleanSelect and Checkbox can return null. That will cause an issue with CoreShop as soon as we introduce PHP 7.1 Type Hints. My idea to solve this would be following:

have following fields:
   - pimcoreActive
   - pimcoreDigitalProduct
   - pimcoreTracked

and implement the methods
   - getActive, setActive
   - getDigitalProduct, setDigitalProduct
   - getTracked, setTracked

and use the Pimcore fields for them. Therefore we can make sure what return value our fields really give.

But, in case of your PR we should be fine.

Ping @solverat What do you think?

dpfaffenbauer:
cheers @SilberMa

Issue #534, opened 2018-07-06
new backend order with variants fails


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Currently it's not possible to create a backend order with variants. Message: `"Product [ID] has no valid TaxRuleGroup"`. Maybe because of the missing `getGetInheritedValues` method.

dpfaffenbauer:
solved

PR #535, opened 2018-07-07
[OrderBundle, CoreBundle] fix Order creation for inherited values

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #534


dpfaffenbauer:
## getDataForObject()
Reason for these two "more or less" equal functions is: EditmodeHelper is used by Embedded Classes, and the one in AbstractSaleController is purely used for loading data from the Sale Object. Once Pimcore changes something in there implementation of what is going on inside the EditmodeHelper, we need to adapt and change that as well. And for the Controller: We would not need to change anything there, cause it does what it should do.

For your error: Can you further investigate why you have issues? I mean, Pimcore does the same thing and it works for me as well.

## Backend UX

###  Currency
New Feature and not part of this PR or the issue linked.

### Confirmation
New Feature and not part of this PR or the issue linked.

### Allow Price Modification
New Feature and not part of this PR or the issue linked.

PR #557, opened 2018-07-25
[All] remove all usages of pimcore_id from container

dpfaffenbauer:
as this causes issues during runtime on first installation where class files don't exist. Change this to be used evaluated on Runtime by the Repositories

| Q             | A
| ------------- | ---
| Bug fix?      | partly
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no

**BC break** As it removes all occurrences of `*.pimcore_id` from container parameters as well as the array of `pimcore_class.ids`

This makes the container build the same with and without generated Pimcore classes.


solverat:
i think, a changelog note is missing.

dpfaffenbauer:
https://github.com/coreshop/CoreShop/commit/dfdbcc799bce5d1860008b3318f06f276721eca9

solverat:
it should be 2.0.0-beta.3 to 2.0.0-beta.4 right? :D

dpfaffenbauer:
https://github.com/coreshop/CoreShop/commit/59ee82db4342b9a4381b74800a78df394c6c2d20

PR #617, opened 2018-08-31
[CoreBundle] Updated the Tracking ItemBuilder to use configured product class 

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes (if you are using a custom product class)
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

The Tracking ItemBuilder used the static CoreShopProduct class. Now the configured product class gets injected.


dpfaffenbauer:
Could you change that to be also compatible when you have multiple Product Classes? So basically, just remove the model-class-name from the constructor and check if the actual $product has the functions.

dpfaffenbauer:
thank you

Issue #619, opened 2018-09-03
[Price Rule] Improve Voucher Price Rule


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | sort of
| BC Break report? | no
| RFC?             | no

## UX Improvements
From business point of view it should be possible to allow multiple voucher codes per cart.

## Code Improvements
Reduce Controller Logic here: 
https://github.com/coreshop/CoreShop/blob/fe780a5e7f4dbadfce44bfbc998c1fed00f7e104/src/CoreShop/Bundle/FrontendBundle/Controller/CartController.php#L58-L97

solverat:
+ Fix _"allow only one voucher cart"_ checkbox in voucher condition: Should work like the label promises.

PR #624, opened 2018-09-04
[CartRules] fix usage of multiple rules with different vouchers per cart

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #619 

This allows to have the same Cart Rule with different voucher codes per cart. This is considered a BC break cause model signature changes -> but this is actually not included in our BC policy.

This is also marked as new feature and Bug fix at the same time, cause the Condition assumed it to be working, but it was never properly implemented to work.


src/CoreShop/Component/Core/Order/Processor/CartRuleAutoRemoverProcessor.php
```diff
@@ -78,7 +80,10 @@ public function process(CartInterface $cart)
 
             $voucherCode = $this->voucherCodeRepository->findByCode($item->getVoucherCode());
 
-            if (!$this->cartPriceRuleValidator->isValidCartRule($cart, $item->getCartPriceRule(), $voucherCode)) {
+            if (!$item->getCartPriceRule()->getIsVoucherRule() && null !== $item->getVoucherCode()) {
```
dpfaffenbauer: this removes rules that are not voucher rules but have a voucher code in the cart

src/CoreShop/Component/Order/Model/CartInterface.php
```diff
@@ -54,19 +54,35 @@ public function hasPriceRules();
     /**
      * @param $priceRule
      */
-    public function addPriceRule($priceRule);
+    public function addPriceRule(ProposalCartPriceRuleItemInterface $priceRule);
```
dpfaffenbauer: this changes the signature to actually fit the expected type -> at least the new expected type of what we store in the Fieldcollection

src/CoreShop/Component/Order/Model/ProposalPriceRuleTrait.php
```diff
@@ -101,23 +114,71 @@ public function removePriceRule($priceRule)
...
+        CartPriceRuleInterface $priceRule,
+        CartPriceRuleVoucherCodeInterface $voucherCode = null
+    ) {
+        return null !== $this->getPriceRuleByCartPriceRule($priceRule, $voucherCode);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriceRuleByCartPriceRule(
```
dpfaffenbauer: this functions check if there is already a Fieldcollection Item for a given set of Cart Rule and Voucher

PR #633, opened 2018-09-13
[NotificationBundle] Fix coreshop_path and inherited product getters …

ktallafus:
…not working in backend context/notification mails (AbstractObject::getGetInheritedValues() is false in backend)

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   

When using coreshop_path/coreshop_url or any product getter in a notification mail template it does not work, if your product is a **variant** and the respective properties get inherited from the paren, because inheritance is disabled in backend context.


dpfaffenbauer:
looks good to me, once the test pass, I'll merge

Issue #643, opened 2018-09-24
[bug] csv import of storePrice: invalid json gets ignored silently


myrdd:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Import a product via csv with `storePrice` being something invalid like `{"1": 0123}` (the integer must not start with a zero), and the storePrice will silently get ignored, because `json_decode` does not throw, but instead returns an empty value.

I'll create a PR.

dpfaffenbauer:
already merged and fixed

PR #644, opened 2018-09-24
[fix] StorePrice CSV import: check for JSON decode errors

myrdd:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #643 

screenshot of the import screen after the fix is applied:

spectacle sl2724

The storePrice value was `{"1":0196}`

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/CoreExtension/StorePrice.php
+++ b/src/CoreShop/Bundle/CoreBundle/CoreExtension/StorePrice.php
@@ -468,6 +468,11 @@ class StorePrice extends Model\DataObject\ClassDefinition\Data
         $storeRepo = $this->getStoreRepository();
 
         $data = $importValue == "" ? [] : json_decode($importValue, true);
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/CoreBundle/CoreExtension/StorePrice.php
```diff
@@ -468,6 +468,11 @@ public function getFromCsvImport($importValue, $object = null, $params = [])
         $storeRepo = $this->getStoreRepository();
 
         $data = $importValue == "" ? [] : json_decode($importValue, true);
+
+        if (json_last_error() !== JSON_ERROR_NONE) {
+            throw new \RuntimeException('Error decoding Store Price JSON `' . $importValue . '`: ' . json_last_error_msg());
```
dpfaffenbauer: Is `\RuntimeException` the right Exception here? I would have though `\InvalidArgumentException` would be better

myrdd: You're right. Too much copy-paste.

dpfaffenbauer:
thanks

PR #664, opened 2018-10-05
[TrackingBundle] fix template notation for tracking

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

LiipTheme introduced a Bug here, so we decided to change the notation, since the colon notation is not supported in Symfony 4 anymore as well.

https://github.com/liip/LiipThemeBundle/issues/194


anyt:
Hi @dpfaffenbauer ,
I have no complaints about bugfix, but here you say
> the colon notation is not supported in Symfony 4 anymore as well.

Could you confirm it with some link to the code or documentation? Can not found it by myself.

PR #672, opened 2018-10-09
Quotes folder is created during installation and Quotes menu works wi…

mgomse:
…thout need to manual addition

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -----

Right now quotes listing menu doesn't work after installation. The reason is that there is no coreshop/quotes folder. This PR fix the installation to add this folder and additionaly add some tiny code improvement so adding/removing folders should be one-line change next time.


dpfaffenbauer:
🎉  thanks a lot

PR #675, opened 2018-10-09
Use decimal as column type instead of double

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Use `decimal` as column type, because `double` is not registered with `\Doctrine\DBAL\Types\Type::addType()`.


dpfaffenbauer:
🎉 thanks :)

PR #676, opened 2018-10-09
Check if value is instance of Carbon already

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Returns the value directly if it already is a instance of `Carbon\Carbon`.


dpfaffenbauer:
thanks a lot

Issue #686, opened 2018-10-23
[SECURITY] User Provider should only search for non-guest customers 


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Scenario
- place order as guest
- register as customer with same credentials as done in the guest-checkout
- logout
- try to login

## Result
- Error: invalid credentials

The ObjectUserProvider does not include the`isGuest` flag so two objects are available, first wins. If the first is a guest object, the login will not work. 

I'll provide a PR for that. I'll also check #662 since it's based on the same issue. We don't need two objects for that, a proper query is enough.

Issue #662, opened 2018-10-04
Do not reset password for guest-accounts


micru:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

There should be a check if a customer has only a guest account before resetting the password. Currently if a customer puts their email into the _forgot password_ form only the existance of a customer with this email is checked. If this email belongs to a guest, the customer gets an reset password mail anyways but will not ne able to login after setting a new password. This is very confusing.

Checking if the customer is guest here https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Bundle/FrontendBundle/Controller/RegisterController.php#L96 and directing back to the _forget password form_ if so, should do the trick for now.

dpfaffenbauer:
are you sure? cause `findCustomerByEmail` checks on if it is a customer and not a guest.

micru:
You are right.. unfortunately we got 2 Customer with the same email. One is a guest, the other not. Looks like this is the real problem. We will investigate this further and might open a new issue if we found the reason why this happened. Guess you might close this issue

dpfaffenbauer:
Yes, we are aware of that and already thought about splitting that up into 2 objects -> User and  Customer

dpfaffenbauer:
superceded by https://github.com/coreshop/CoreShop/issues/665

PR #687, opened 2018-10-23
[CoreBundle] check guest state in ObjectUserProvider

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #662, #686 

This PR will fix login issues described in the issues above.

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Security/ObjectUserProvider.php
+++ b/src/CoreShop/Bundle/CoreBundle/Security/ObjectUserProvider.php
@@ -27,11 +27,18 @@ class ObjectUserProvider implements UserProviderInterface
     protected $customerRepository;
 
     /**
-     * @param CustomerRepositoryInterface $customerRepository
+     * @var string
      */
-    public function __construct(CustomerRepositoryInterface $customerRepository)
+    protected $className;
+
+    /**
+     * @param CustomerRepositoryInterface $customerRepository
+     * @param string                      $className
+     */
+    public function __construct(CustomerRepositoryInterface $customerRepository, $className)
     {
         $this->customerRepository = $customerRepository;
+        $this->className = $className;
     }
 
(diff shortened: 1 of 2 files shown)


src/CoreShop/Bundle/CoreBundle/Security/ObjectUserProvider.php
```diff
@@ -67,10 +69,32 @@ public function loadUserByUsername($username)
     {
         $this->checkClass();
 
-        $getter = sprintf('getBy%s', ucfirst($this->usernameField));
+        /** @var Concrete $class */
```
dpfaffenbauer: Shouldn’t we better inject the repo?

solverat: I'm not sure about this since this class is a very generic one, see comment on the`className` property:

```
The pimcore class name to be used. 
Needs to be a fully qualified class name
(e.g. Pimcore\Model\DataObject\User or your custom user class
extending the generated one.)
```

dpfaffenbauer: Just checked again, and we should change that to the Repository. The className variable is always just the Customer Class -> so the repo uses the same class. And therefore we can just use the Repo function which does exactly the same.

dpfaffenbauer:
Why not use findCustomerByEmail function?

solverat:
Because it could be any field.

dpfaffenbauer:
yes currently, but since the parameter for username is hardcoded anyway, we can change that to use the fixed coreshop properties for that. I don't see the point of having that configureable, I think I just took the Pimcore example and did the same. Can we refactor that?

Issue #724, opened 2018-11-22
Shipment Tax Rate Summary ignores floating Numbers


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

- Create a tax rule with `7.7%` rate
- Create shipment, assign tax rule
- Create order
=> Open Order Object: Shipment tax rate in numbers section is wrong (`7`)

PR #725, opened 2018-11-22
[CoreBundle, Pimcore] Fix for Shipping Tax Rate

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #724

### DIFF
--- a/features/order/order_shipping.feature
+++ b/features/order/order_shipping.feature
@@ -1,0 +1,32 @@
+@order @order_shipping
+Feature: Create a new order
+  And test if the shipping is calculated right
+
+  Background:
+    Given the site operates on a store in "Austria"
+    And the site has a currency "Swiss Franc" with iso "CHF"
+    And the site operates on locale "en"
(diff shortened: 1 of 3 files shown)


src/CoreShop/Behat/Context/Transform/CartContext.php
```diff
@@ -25,26 +26,46 @@
...
      */
     public function cart()
     {
         return $this->cartContext->getCart();
     }
+
+    /**
+     * @Transform /^loaded cart(?:|s)/
+     */
+    public function loadedCart()
```
dpfaffenbauer: just to be clear here, the `loaded cart` is to test if the value gets persisted right. If we don't do that, we can set any value we want and the test would pass. even string values would be accepted by Pimcore. So, once we force load the object, we get the newly hydrated object and can test against that.

PR #730, opened 2018-11-26
Country (mainly) and currency (additionaly) fixtures fixes

mgomse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   

Fixes for command `coreshop:fixture:data:load`:
* Country:
  * **major fix** - countries weren't correctly installed for locals like en_GB of fr_FR
  * minor fix/upgrade - if a country was already active (in case of a 2nd run) leave it active
* Currency:
  * similar to Country Fixture - if Currency is already there don't add it a 2nd time (in case of run second time the command and remove a row from coreshop_fixtures_data of course)


dpfaffenbauer:
👍  thanks a lot @mgomse 

will merge once the tests are all green

PR #731, opened 2018-11-26
Fix for carriers add button

mgomse:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

Main fix is in `/src/CoreShop/Bundle/ShippingBundle/Resources/public/pimcore/js/carrier/panel.js` - couriers add button wasn't working properly (identifier wasn't filled after add). 

Additionaly according to the boy scout rule I also added some fixes for scrutinizer (mainly docs).

Fixtures changes are from https://github.com/coreshop/CoreShop/pull/730 (I forgot to create new branch from master, but I hope it won't be a problem after a merge)

### DIFF
--- a/src/CoreShop/Behat/Context/Setup/IndexContext.php
+++ b/src/CoreShop/Behat/Context/Setup/IndexContext.php
@@ -145,7 +145,6 @@ final class IndexContext implements Context
             $index->addColumn($column);
 
             $this->objectManager->persist($column);
-            // $row['name'], $row['email'], $row['phone']
         }
 
         $this->saveIndex($index);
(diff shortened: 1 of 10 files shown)


dpfaffenbauer:
you are completely mix and matching different things with this PR. Could you please create separate ones for each thing you wanna fix.

Also: Regarding the wrong branch -> you can always rebase your branch onto another branch and force-push it here. So you end up with the right history.

I will close this PR untouched and I hope you'll create a new one, or multiples.

PR #735, opened 2018-12-01
Remove login customer after successfully registration

kwreczycki:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Hi, 
I think that logging user after successfully registration process is not always needed case in business requirement.

Registration manager should just register new customer and all additional actions should be attached to customer.register event.

Please let me know what you think about that approach.


dpfaffenbauer:
@kwreczycki this still seems not the be finished. You also need to register the event-listener. Otherwise this would be a BC break as it changes behaviour. For your use-case: you can overwrite either the listener or the `UserLoginService` to not get logged in.

kwreczycki:
@dpfaffenbauer Yep, I know, that's why i tagged this PR by WIP. 
I should finish it in a few days...

dpfaffenbauer:
Looks all good to me, but I have to test it first.

dpfaffenbauer:
thanks a lot

kwreczycki:
:+1:

PR #740, opened 2018-12-05
Fix for StoreMailActionProcessor exception in Notification Rule.

adrozdek:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When I add 'Store based Mail' action in 'User Reset Password' I get the error during reset password request (dispatch event 'coreshop.customer.request_password_reset'):
`'StoreMailActionProcessor: Store is not set.'`
This PR should fix this.

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/EventListener/NotificationRules/CustomerListener.php
+++ b/src/CoreShop/Bundle/CoreBundle/EventListener/NotificationRules/CustomerListener.php
@@ -164,6 +164,7 @@ final class CustomerListener extends AbstractNotificationRuleListener
             'lastname' => $customer->getLastname(),
             'email' => $customer->getEmail(),
             'object' => $customer,
+            'store' => $this->shopperContext->getStore(),
         ];
     }
 }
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/CoreBundle/EventListener/NotificationRules/CustomerListener.php
```diff
@@ -164,6 +164,7 @@ private function prepareCustomerParameters(CustomerInterface $customer)
             'lastname' => $customer->getLastname(),
             'email' => $customer->getEmail(),
             'object' => $customer,
+            'store' => $this->shopperContext->getStore(),
```
dpfaffenbauer: I think we should check if we even have a shop here with
```
$this->shopperContext->hasStore()
```

@solverat What do think?

adrozdek: @dpfaffenbauer I didn't notice that getStore throws an exception. I changed it to hasStore->getStore version.

dpfaffenbauer:
thanks a lot for your first PR @adrozdek 

🎉 🍻

PR #750, opened 2018-12-13
Add logout to CustomerLogin service

kwreczycki:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #...   

Hi Dominik,
I think that alternatively to loginCustomer in CustomerLoginService we should have also possibility to logout him.
What do you think about that ?

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginService.php
+++ b/src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginService.php
@@ -15,6 +15,8 @@ namespace CoreShop\Bundle\CoreBundle\Customer;
 use CoreShop\Component\Core\Model\CustomerInterface;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorage;
 use Symfony\Component\Security\Core\Authentication\Token\UsernamePasswordToken;
+use Symfony\Component\HttpFoundation\RequestStack;
+
 
 final class CustomerLoginService implements CustomerLoginServiceInterface
 {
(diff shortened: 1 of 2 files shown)


src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginServiceInterface.php
```diff
@@ -20,4 +20,6 @@
      * @param CustomerInterface $customer
      */
     public function loginCustomer(CustomerInterface $customer);
+
+    public function logoutCustomer();
```
dpfaffenbauer: missing phpdoc

src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginService.php
```diff
@@ -36,4 +44,10 @@ public function loginCustomer(CustomerInterface $customer)
         $token = new UsernamePasswordToken($customer, null, 'coreshop_frontend', $customer->getRoles());
         $this->securityTokenStorage->setToken($token);
     }
+
+    public function logoutCustomer()
```
dpfaffenbauer: missing phpdoc

src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginService.php
```diff
@@ -36,4 +44,10 @@ public function loginCustomer(CustomerInterface $customer)
...
+    {
+        $this->securityTokenStorage->setToken(null);
+        $this->requestStack->getCurrentRequest()->getSession()->invalidate();
```
dpfaffenbauer: I am not sure about that. Why do you even need that anyway? Couldn't you just call the logout url directly?

dpfaffenbauer: I mean, that is something the symfony firewall handles

dpfaffenbauer:
Also, this PR targets the 2.0 release, can you rebase to 2.0 pls?

src/CoreShop/Bundle/CoreBundle/Customer/CustomerLoginService.php
kwreczycki: Yeah, maybe you have right. I wanted add this method only to have consistent behaviour for this service. If we have login we expect logout also but in fact it's bit overengineer.

dpfaffenbauer:
I guess we can close that.

PR #751, opened 2018-12-19
WIP: Exiting early in SessionAndStoreBasedContext in order not to start a …

khusseini:
…session if has not already been started

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

In one of our projects we had to minimize the sections where a session is started in order to take full advantage of pimcore's page caching. I added a few lines to the `SessionAndStoreBasedContext` to exit early if a session hasn't already been started.


khusseini:
I just noticed that there are a few places, however I will have to test this through, because it seems it might break the cart. I set the title to WIP for now until I am sure there are no side-effects

khusseini:
This is going to be a bigger construction site. I'll close the PR for now

dpfaffenbauer:
why? Whats the issue?

PR #754, opened 2019-01-02
[All] require 2.1 from packages

dpfaffenbauer:


dpfaffenbauer:
nope, we are not gonna do that yet. It is not necessary and will cause issues when you want to include packages from 2.0 and 2.1

PR #761, opened 2019-01-04
[Core] check if $categories is array before checking count in Tracking Extractors

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Fixes a common PHP7.2 issue with the `count()` method and `null` variables


dpfaffenbauer:
pls check the PR, you include a lot of commits that don't belong here.

dpfaffenbauer:
I'll close this for now, cause it is useless

PR #762, opened 2019-01-07
[Travis] fix tests for Pimcore 5.6

dpfaffenbauer:


dpfaffenbauer:
this will fail, cause Pimcore 5.6 introduced some breaks

PR #767, opened 2019-01-10
[Core] fix bugs in Tracking Extractors 

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

check if variable is an array before checking the count


dpfaffenbauer:
thanks

PR #769, opened 2019-01-13
add lost docu part - how to extend own Frontend-Controllers

ITspirit:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Bring lost docu (https://www.coreshop.org/docs/2.0.0/Development/Override_Controllers.html) back into new doc structure.

### DIFF
--- a/docs/03_Development/15_Store_Front/01_Controllers.md
+++ b/docs/03_Development/15_Store_Front/01_Controllers.md
@@ -50,3 +50,17 @@ core_shop_frontend:
     controllers:
         product: AppBundle\Controller\ProductController
 ```
+
+**3**: CoreShop Override Controller (optional)
+
+CoreShop uses services for controllers, if you need to extend a controller, simply override the service:
+
+```yaml
+services:
+    coreshop.frontend.controller.category:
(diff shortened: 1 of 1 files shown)


ITspirit:
Well, that´s true if my "own" controller doesn´t have any other in the constructor, but if the constructor has some more parameters, your approch won´t work anymore and the service.yml must be extended too.

dpfaffenbauer:
Thats true, but overwriting a service will never be BC compatible. It works in Symfony, but once I change service names, it will break in your installation.

What I am saying with it: It works, but I don't recommend it and also it is something specific to symfony and not CoreShop, thats why I will not merge it.

ITspirit:
Well, ok, but it would be the same if you choose to rename the Controller. Also a BC.
For this cases we use semantic versioning and documentation in changelogs (normally)

Can´t see any other (better) approach to extend the Controllers (with additional constructor parts) at the moment.

If you know one - please let me know. Thx.

dpfaffenbauer:
You are right about the BC.

Don't use the CoreShop ones, instead create your own. Thats the best solution.

PR #809, opened 2019-02-07
[Pimcore] add object_method twig function

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Allows to check if object method exists. Native twig implementation `{{ attribute(object, method) is defined }}` sometimes fails and returns `true`, although the method doesn't exist.


dpfaffenbauer:
@aarongerig can you rebase to 2.0 pls?

dpfaffenbauer:
thanks 🎉

Issue #665, opened 2018-10-05
Split User and Customer into two Objects.


dpfaffenbauer:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | yes
| RFC?             | yes

With the current implementation, it is quite hard to decide and understand what user is a customer and what "just" a guest customer, in order to make that better dividable, split the entities into two.

dpfaffenbauer:
@solverat this should also make the Members Integration easier right?

dpfaffenbauer:
I disagree. In terms of separating concerns, it should be divided. The customer is the person who actually buys. The user is the entity that we authenticate against. Currently we use the same for both.

Current solution could also raise issues with CMF, cause a guest user is no a customer per se. And we store them to the CMF currently.

dpfaffenbauer:
Well, at least not a recurring customer we can perform CMF actions against

Cruiser13:
Upvoting this.

dpfaffenbauer:
will not happen for now. as discussed internally

Cruiser13:
Too bad. Would have brought a lot of benefits in my opinion. If it is not planned for now but not off the table - maybe leave it open and work on this at a later point?

dpfaffenbauer:
we had a discussion on the weekend that the benefits aren't that great to bring that big of a bc break. what would be your benefits?

Cruiser13:
Here are some:

- allow browsing store without cookies (DSGVO compliant)
- allow customers (for example B2B) without accounts
- allow registrations (for a free service for example) without customer details such as customer ID
- allow other kind of customers (for example one company customer with 3 different employees each having a user account but using the same customer account)
- easy implementation of different members and customer classes. If you setup a coreshop customer implementing members and CMF the class is extremly huge and hard to handle atm - also slows down the system

In my opinion they are just two separate data models which should kept separately therefor.

dpfaffenbauer:
1. AFAIK: CoreShop doesn't trigger a session until the customer is logged in, so there would be no change
2. Thats already possible using the "isGuest" Flag. But true, I would prefer the other solution to solve this more
3. You still need to have a Customer to put future order against, so thats not valid
4. True
5. True

@solverat What is your opinion about this arguments?

Cruiser13:
[...]

Issue #186, opened 2017-09-13
Implement MembersBundle for Customer Context and Frontend User Management


dpfaffenbauer:
https://github.com/dachcom-digital/pimcore-members

Install MembersBundle with CoreShop to take advantage of this amazing Bundle. This also enables more possibilities for authentication stuff when using CoreShop. For example:
 - Single Auth for your Shop, as well as your Forum/Blog/Website
 - Restrict certain Products/Categories to Members and let Members take care about restricting
 - Restrict certain Documents to Members and let Members take care about restricting

How To:
 - ~~leave CustomerComponent and CustomerBundle as it is~~
 - ~~leave CoreComponent and CoreBundle as it is~~
 - ~~implementation is done in MembersBridge (CoreShop\Bridge\MembersBundle)~~
 - ~~MembersBundleBridge can be activated manually if needed~~
 - add docs on what is necessary to use MembersBundleBridge
 - ~~implement MembersBundle\Adapter\User\AbstractUser in Bridge (a trait would make life easier)~~
 - ~~implement MembersBundle\Adapter\Group\AbstractGroup in Bridge (a trait would make life easier)~~
 - ~~change ProductRepository and CategoryRepository and add MembersQueries~~
 - use MembersRegistrationFormType, MembersLoginFormType, MembersForgotPasswordFormType (really necessary? can't we stick with our FormTypes?)
 - ignore members emails, is done via NotificationRules in CoreShop (maybe map Members Mails to CoreShop Customer Notification Types)

@solverat I guess I need some help here and probably some changes in MembersBundle:
 - could you create a trait for what is currently done in MembersBundle\Adapter\User\AbstractUser and MembersBundle\Adapter\Group\AbstractGroup? So I can use your trait and just implement the UserInterface and GroupInterface
 - can you check if my How-To makes sense?

dpfaffenbauer:
@solverat need some help here from you, currently this is working as Bridge and or Core (not sure which way to go):

 - Restriction of Products and Categories
 - Customer and CustomerGroup is used as Member User and Member User Group

What I am not sure of: Should CoreShop use Members Events and Controllers as well? Or should it just stick to what its currently already doing?

solverat:
CoreShop should use Members in Core not as Bridge. We'll have to discuss further steps in our next hackathon. ;)

dpfaffenbauer:
closed due to too old and not sure when, how and what actually.

dpfaffenbauer:
[...]

PR #823, opened 2019-02-19
[WIP] split customer and user into seperate entities

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | yes
| Fixed tickets | #665


dpfaffenbauer:
This potentially makes the Members Integration easier as well (https://github.com/coreshop/CoreShop/issues/186)

solverat:
I still don't get it... What are the real benefits here?

dpfaffenbauer:
- A guest is a customer, but not a user. A guest currently has ROLES and is logged into the firewall.
 - Integration with something like Members is now done with the Users Component/Bundle, so we don't need to care about Guests as well
 - You can convert a guest customer to a full customer by just adding a user to it

dpfaffenbauer:
will not happen for now. as discussed internally

PR #850, opened 2019-03-04
[All] Autowiring

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

make more use of auto-wiring in symfony 3.4. This will automatically apply tags to classes implementing the needed interface for all kinds of Rule Conditions/Actions, Price Calculators, Index/Filters, Contextes. This also allows auto-wiring of coreshop generated services like repositories and manager by using the name of it. Like productRepository` or `storeRepository`.

If a tag is not supplied, but the compiler-pass expects a `type`, the `type` is getting auto-generated by the class-name.


dpfaffenbauer:
rebased to master -> will be released with 2.1

PR #867, opened 2019-03-15
Make Embedded Class compatible with Pimcore 5.6

sebsasto:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

The Embedded Classes were not loading because some compatibility errors. Specifically, the `$typename` was missing, a circular reference on the preGet method and the `EditmodeHelper` wasn't populating its attributes as expected.


dpfaffenbauer:
FYI: Your PR is fine, tests fail cause Pimcore introduced a lot of BC breaks again

dpfaffenbauer:
thanks a lot

Issue #837, opened 2019-02-25
Sale Unit


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Sale Units should be possible during product configuration:

## Product Settings
- Allow Additional Units (Each additional unit requires a dedicated store price)
- Default: "Item" (After Migration)

![bildschirmfoto 2019-02-25 um 16 25 31](https://user-images.githubusercontent.com/700119/53347727-01953780-391a-11e9-81e0-e23224f19ceb.png)

## Unit Definition
- Add a panel under `CoreShop` -> `Products` -> `Product Units`

![bildschirmfoto 2019-02-25 um 15 29 42](https://user-images.githubusercontent.com/700119/53344227-795f6400-3912-11e9-8650-de5a56161c76.png)

## Quantity Pricing
- Allow volume pricing rules based on different sale units
- "Item" would be the default value

![bildschirmfoto 2019-02-25 um 15 34 43](https://user-images.githubusercontent.com/700119/53344400-e4a93600-3912-11e9-86f7-57860b91aba0.png)

PR #875, opened 2019-03-19
Sale Unit Follow Up

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #837


src/CoreShop/Bundle/CoreBundle/Doctrine/ORM/ProductStoreValuesRepository.php
```diff
@@ -44,7 +44,7 @@ public function findForProductAndStore(ProductInterface $product, StoreInterface
             ->setParameter('product', $product->getId())
             ->setParameter('store', $store)
             ->getQuery()
-            ->useResultCache(true)
+            ->useResultCache(false)
```
dpfaffenbauer: completely remove instead of setting false

PR #899, opened 2019-03-25
[Core] don't overwrite interfaces

dpfaffenbauer:
as this breaks how stacks are supposed to work. And it also doesn't make any sense

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


dpfaffenbauer:
I have to revert this, as it breaks more than it fixed.

PR #918, opened 2019-04-02
Clone AdditionalData Line By Line

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

`$sale->setAdditionalData($cart->getAdditionalData());` is evil since it transfers:

`Pimcore\Model\DataObject\CoreShopCart\AdditionalData` into the `CoreShopOrder` Object instead of using the `Pimcore\Model\DataObject\CoreShopOrder\AdditionalData` ObjectBrick Class.

### DIFF
--- a/src/CoreShop/Component/Core/Order/Transformer/CartToSaleTransformer.php
+++ b/src/CoreShop/Component/Core/Order/Transformer/CartToSaleTransformer.php
@@ -20,6 +20,7 @@ use CoreShop\Component\Order\Model\ProposalInterface;
 use CoreShop\Component\Order\Model\SaleInterface;
 use CoreShop\Component\Order\Transformer\ProposalTransformerInterface;
 use CoreShop\Component\Payment\Model\PaymentSettingsAwareInterface;
+use Pimcore\Model\DataObject\Objectbrick;
 use Webmozart\Assert\Assert;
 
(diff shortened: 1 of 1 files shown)


src/CoreShop/Component/Core/Order/Transformer/CartToSaleTransformer.php
```diff
@@ -68,7 +69,23 @@ public function transform(ProposalInterface $cart, ProposalInterface $sale)
                 $sale->setShippingTaxRate(0);
             }
 
-            $sale->setAdditionalData($cart->getAdditionalData());
+            $cartAdditionalData = $cart->getAdditionalData();
+            $saleAdditionalData = $sale->getAdditionalData();
```
dpfaffenbauer: since the sale is actually new here, what if it returns null? than the data is never transfered

solverat: That's not possible: 
image

dpfaffenbauer: actually it is: if the class is missing right? I think we should just double check to prevent issues on order creation.

PR #922, opened 2019-04-03
prevents recursion if href-elements contain entities of Class CoreShopProduct

smaier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

prevents recursion if href-elements contain entities of Class CoreShopProduct.


dpfaffenbauer:
1. please rebase on 2.0 branch
2. I don't feel like thats a definite solution, since this can also happen with other entities. Better to introduce a recursion detection.

dpfaffenbauer:
actually, this is completly wrong anyway, since CoreShop itself doesn't really know `DataObject\CoreShopProduct`

solverat:
Also: Could be any other product class eg. `DataObject\MySpecialProduct` or even mlutiple product classes.

smaier:
so `instanceof \CoreShop\Component\Core\Model\Product` would be the way to go?

PR #926, opened 2019-04-03
Feature/order detail recursion

smaier:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

tries to prevent recursions for associated product-entities


smaier:
didn't know how to change my pull-request as i had to change the base-branch

dpfaffenbauer:
ok, don't use models directly ;)

Also, as said in the other ticket, I am not happy with that, since a recursion could happen with every kind of object. Can you create a ticket instead? I'll sort that differently

PR #931, opened 2019-04-08
Specify how to get latest stable version

Cruiser13:
There are stable releases now which should probably be preferred for productive systems.


dpfaffenbauer:
Actually, that should go into the 2.0 branch and merged into master from there. So, can you rebase that on 2.0 pls?

Issue #937, opened 2019-04-10
Show a default for non translated Countries


kmpm:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

When using a non English language in admin it's quite difficult to administer countries.
In the list where you select country all non translated countries show as empty string.
Default fallback to English or ISO code would be great if possible.
![Language Translation](https://user-images.githubusercontent.com/96618/55864343-97102200-5b7c-11e9-9a40-b39c0590c677.png)

dpfaffenbauer:
👍

dpfaffenbauer:
fallsback now to isoCode

PR #944, opened 2019-04-10
[AddressBundle] follow-up on #939

dpfaffenbauer:
change it on the server as well, so we also display the fallback for other country field types as well

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #937


dpfaffenbauer:
Bildschirmfoto 2019-04-10 um 20 43 42
Bildschirmfoto 2019-04-10 um 20 43 47

PR #967, opened 2019-05-02
[Migration] only add indices to store_price table if table actually e…

dpfaffenbauer:
…xists. In some rare cases, the table might not exist anymore

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


dpfaffenbauer:
tests also fail due to #966

PR #991, opened 2019-05-16
[WIP] refactor pimcore object data-tags to use DI for dependencies

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes

This PR takes use of the Pimcore Implementation Loader to load Custom Object Data Tags. This reduces the usage of `Pimcore::getContainer` and thus the risk of using private services.

Since Pimcore doesn't really 100% support this, it is a bit tricky. We cannot store properties directly in the Data Class, since Pimcore uses `var_export` to store definition files, this also exports the dependencies and that will fail with the entity manager. So, we use a private static variable to handle that.

### DIFF
--- a/src/CoreShop/Bundle/AddressBundle/CoreExtension/AddressIdentifier.php
+++ b/src/CoreShop/Bundle/AddressBundle/CoreExtension/AddressIdentifier.php
@@ -12,30 +12,6 @@
 
 namespace CoreShop\Bundle\AddressBundle\CoreExtension;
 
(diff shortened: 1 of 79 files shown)


dpfaffenbauer:
this will not merged until this is resolved: https://github.com/pimcore/pimcore/issues/4390

currently it is just to hot and could cause potential issues.

PR #1004, opened 2019-05-28
always load product unit definitions

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | ---

Do not lazy load `defaultUnitDefinition` and `unitDefinitions`  in `ProductUnitDefinitions`. Otherwise the pimcore version preview will throw an exception.


dpfaffenbauer:
thanks

Issue #1075, opened 2019-08-26
Add Precision to Unit Definition


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Currently, we cannot add a precision to units of quantity like `2.50` kg. Adding a precision field to the default and every additional unit should fix this.

PR #1081, opened 2019-09-02
Unit Definition - Precision

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1075


solverat:
###  This PR allows to:

- Add `precision` to each unit of measure
- Allow adding quantity with decimals

### This PR changed:
- Use `float` instead of `int` in every quantity workflow
 

image

Kapture 2019-09-03 at 14 35 59

## Frontend

Kapture 2019-09-03 at 16 08 04

## BC Breaks
Quantity-Field is now a "text" field and needs to be watched by a javascript helper. There is also a serverside validation.

CHANGELOG-2.1.x.md
```diff
@@ -1,6 +1,9 @@
 # Within 2.1
 
 ## 2.1.0
+ - BC-Break: `quantity` Field is now a `text` field instead of `number`:
```
dpfaffenbauer: what quantity field?

src/CoreShop/Component/Core/Order/Modifier/CartItemQuantityModifier.php
```diff
@@ -13,25 +13,90 @@
...
+        }
+
+        $product = $item->getProduct();
+        if (!$product instanceof ProductInterface) {
+            return $targetQuantity;
+        }
+
+        $scale = $this->getScale($item);
+        if ($scale === null) {
+            return $targetQuantity;
```
dpfaffenbauer: since it has no scale here, shouldn't we remove possible floating numbers?

CHANGELOG-2.1.x.md
solverat: The frontend quantity field in the addToCartForm and within the cartForm

src/CoreShop/Component/Core/Order/Modifier/CartItemQuantityModifier.php
solverat: the return value of the `getScale` method can be `null` or `integer`, no floating numbers are possible here.

CHANGELOG-2.1.x.md
dpfaffenbauer: why not a number that allows precision?

solverat: https://caniuse.com/#feat=input-number

Mobile:
- UI widget does not include increment/decrement buttons.
- UI widget does not take the "step", "min" or "max" attributes into account.

solverat: changed back to number field

dpfaffenbauer:
thanks a lot, again.

PR #1089, opened 2019-09-05
Protected members are available only via getters.

rishadomar:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #...

### DIFF
--- a/src/CoreShop/Bundle/PimcoreBundle/CoreExtension/EmbeddedClass.php
+++ b/src/CoreShop/Bundle/PimcoreBundle/CoreExtension/EmbeddedClass.php
@@ -57,6 +57,7 @@ final class EmbeddedClass extends Multihref
 
         $returnData = [];
 
+        $i = 0;
         foreach ($data as $embeddedObject) {
             if (!$embeddedObject instanceof DataObject\Concrete) {
                 continue;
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/PimcoreBundle/CoreExtension/EmbeddedClass.php
```diff
@@ -72,12 +73,12 @@ public function getDataForEditmode($data, $object = null, $params = [])
             list('objectData' => $objectData['data'], 'metaData' => $objectData['metaData']) = $editmodeHelper->getDataForObject($embeddedObject);
 
             $objectData['id'] = $embeddedObject->getId();
+            $allowedKeys = ['o_published', 'o_key', 'o_id', 'o_modificationDate', 'o_creationDate', 'o_classId', 'o_locked', 'o_type', 'o_parentId', 'o_userOwner', 'o_userModification'];
```
dpfaffenbauer: I am not really sure how that should solve the problem.

why not simply like this:

```php
$objectData['general'] = [
                'index' => $embeddedObject->getIndex(),
                'o_published' => $embeddedObject->getPublished(),
                'o_key' => $embeddedObject->getKey(),
                'o_id' => $embeddedObject->getId(),
                'o_modificationDate' => $embeddedObject->getModificationDate(),
                'o_creationDate' => $embeddedObject->getCreationDate(),
                'o_classId' => $embeddedObject->getClassId(),
                'o_className' => $embeddedObject->getClassName(),
                'o_locked' => $embeddedObject->getLocked(),
                'o_type' => $embeddedObject->getType(),
                'o_parentId' => $embeddedObject->getParentId(),
                'o_userOwner' => $embeddedObject->getUserOwner(),
                'o_userModification' => $embeddedObject->getUserModification()
            ];
```

rishadomar: Yes sure - much neater! 
It worked because the other members were public but I like using the getters instead.
Also, getIndex() returns 0 unless you save and reload the object. But if you close the object and view again it's reverted to 0.
So I added the ++$i.
(Also users prefer index to start with 1 instead of 0)

dpfaffenbauer:
thanks

PR #1094, opened 2019-09-11
[FrontendBundle]: Fix configuration key

khusseini:
Corrected the config tree's root node name in order to make configuration available to app

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #...   

Fixed configuration root key. Currently it is not possible to configure the frontend-bundle without this fix.

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/DependencyInjection/Configuration.php
+++ b/src/CoreShop/Bundle/FrontendBundle/DependencyInjection/Configuration.php
@@ -39,7 +39,7 @@ final class Configuration implements ConfigurationInterface
     public function getConfigTreeBuilder()
     {
         $treeBuilder = new TreeBuilder();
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
are you sure? the root key usually is defined by symfony.

Anyhow, please rebase to 2.0, since that version is also affected.

solverat:
Yes, it's a bit confusing though since the root node is calling `coreshop_frontend`, in a project configuration its `core_shop_frontend`. I also stumbled on this behavior. 

However, this works perfectly fine in my current dev-master project:

```yml

core_shop_frontend:
    controllers:
        category: AppBundle\Controller\CategoryController
```

It's maybe a good idea to change all root node identifier...

PR #1100, opened 2019-09-13
Fixing FrontendBundle Configuration

khusseini:
Corrected the config tree's root node name in order to make configuration available to app

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #...   

Fixed configuration root key. Currently it is not possible to configure the frontend-bundle without this fix.


dpfaffenbauer:
I've changed all the others as well, either we change all of them, or none of them.

Honestly, I can't see any issue with the "wrong" identifier, since it is overwritten by symfony anyway, and I don't experience any issues in my project neither.

dpfaffenbauer:
thanks

Issue #1104, opened 2019-09-13
[FrontendBundle]: CategoryController hardcoded allowed per page


khusseini:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The `perPageAllowed` view variable in `CategoryController::indexAction` is set to a hardcoded array, although a configuration is available. 

https://github.com/coreshop/CoreShop/blob/2.0/src/CoreShop/Bundle/FrontendBundle/Controller/CategoryController.php#L198

PR #1105, opened 2019-09-13
CategoryController perPage Configuration fix

khusseini:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1104 

Set `CategoryController::index`'s `perPageAllowed` view variable to content from configuration rather than a hardcoded value


dpfaffenbauer:
thanks

Issue #1157, opened 2019-10-31
[QuantityPriceRules] Workflow to allow to make inherited QPR on variants independent 


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

## Implementation
Add Button to Main-Product to dispatch Workflow. This will improve #1137.

## Workflow Pre-Checks
- Check if product has any variants
- Check if product has unit definitions
- Add "are you sure" with some explanation about what's going to happen

## Workflow
- Copy and save unit definition to each variant
- Copy all QPR and replace inherited unit with the copied one

![image](https://user-images.githubusercontent.com/700119/67945942-77a41980-fbe0-11e9-861e-08460af4731e.png)

![image](https://user-images.githubusercontent.com/700119/68119185-543fde00-ff02-11e9-98e8-83d2f412f05f.png)

![image](https://user-images.githubusercontent.com/700119/68119350-cc0e0880-ff02-11e9-86a5-2069500a3047.png)

solverat:
Done.

PR #1158, opened 2019-11-04
Implement Variant Unit and QPR Solidifier

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1157


solverat:
image

Workflow:
- Pre-Check and inform user about what is going to happen
- If product has unit definitions, clone them to every `getChildren` of type `variant`
- If product has QPR, clone them
- If product has QPR with unit definitions, search for same unit name in new object (otherwise the unit definition stays null => defined by __clone)

PR #1190, opened 2019-11-25
Fix comment about variants

Cruiser13:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | no


dpfaffenbauer:
Will not be merged, the function should be removed, it is unrecommended to use it. It is untested, very old and not used in the core anyway. so we recommend to not use it.

PR #1192, opened 2019-11-28
[OrderBundle] add sale-detail event

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

### DIFF
--- a/src/CoreShop/Bundle/OrderBundle/Controller/AbstractSaleDetailController.php
+++ b/src/CoreShop/Bundle/OrderBundle/Controller/AbstractSaleDetailController.php
@@ -12,6 +12,7 @@
 
 namespace CoreShop\Bundle\OrderBundle\Controller;
 
+use CoreShop\Bundle\OrderBundle\Events;
 use CoreShop\Component\Address\Formatter\AddressFormatterInterface;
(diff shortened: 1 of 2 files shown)


src/CoreShop/Bundle/OrderBundle/Controller/AbstractSaleDetailController.php
```diff
@@ -278,6 +279,10 @@ protected function getDetails(SaleInterface $sale)
             $jsonSale['priceRule'] = $rules;
         }
 
+        $event = new GenericEvent($sale, $jsonSale);
+
+        $this->get('event_dispatcher')->dispatch('coreshop.sale.detail.prepare', $event);
```
solverat: - Adding `coreshop.sale.detail.prepare` as constant maybe?
- Passing `$jsonSale` as argument is not enough if guess (no reference). Adding a `$jsonSale = $event->getArguments()` after firing the event should do the job.

dpfaffenbauer: done

PR #1194, opened 2019-11-30
Fix Character Length Count in Voucher Code Generator

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

`33` is a defined length for `FORMAT_ALPHANUMERIC` - if you're using `FORMAT_ALPHABETIC` or `FORMAT_NUMERIC` only, it will return null values which generates codes with wrong lengths.


dpfaffenbauer:
thanks

PR #1199, opened 2019-12-15
Add info about coreShopStore fieldtype

Cruiser13:
Propose to add info about working with the coreShopStore fieldtype because it differs from pimcores many-to-types. Docs only.

### DIFF
--- a/docs/03_Development/11_Stores/01_CRUD.md
+++ b/docs/03_Development/11_Stores/01_CRUD.md
@@ -60,3 +60,14 @@ $store = $storeRepository->findById(1);
 $container->get('coreshop.manager.store')->remove($store);
 $container->get('coreshop.manager.store')->flush();
 ```
+
+## Working with stores and pimcore objects
+Note that if you are working with pimcore objects and the coreShopStore fieldtype you can not set the store but instead you have to set the ID of the store.
+```php
+// Fetch Store
+
+$store = $storeRepository->findById(1);
+$customer = \Pimcore\Model\DataObject\CoreShopCustomer::getById(2);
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
I am pretty sure you can do both, the ID and the Store object itself

Cruiser13:
That's actually not the case (took us hours to debug it in our special situation) so I thought we might add it to the docs.

dpfaffenbauer:
are you sure? CoreShop uses that all over the place....

Except is the coreShopStore**s** type, where it only accepts ID's

Cruiser13:
Yes. The following does not work:

```php
public function changeStoreAction(ShopperContext $shopperContext, StoreRepositoryInterface $storeRepository){
$customer = $shopperContext->getCustomer();
$store = $storeRepository->find(2);
$customer->setCurrentstore($store);
}
```
This does:
```php
public function changeStoreAction(ShopperContext $shopperContext, StoreRepositoryInterface $storeRepository){
$customer = $shopperContext->getCustomer();
$store = $storeRepository->find(2);
$customer->setCurrentstore($store->getId());
}
```

Customer class does hold a field of type coreShopStore with the id currentstore.

dpfaffenbauer:
I can't confirm that, I tried following:

```php
$store = $this->container->get('coreshop.repository.store')->findStandard();

        $test = new Test2();
        $test->setName("blub");
        $test->setStore($store);
        $test->setParentId(1);
        $test->setKey('test');
        $test->save();
```

Where store is a CoreShop Store Select Field.

Cruiser13:
Very strange, it does not work for me with my code. Nevermind in this case.

PR #1201, opened 2019-12-16
Allow sales numbers to be more dynamic

Cruiser13:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This would give us the possibility to use numbers like:
O2019-12-12345678
by prefix config:
O{year}-{month}-

Similar to the shopware feature: https://docs.shopware.com/en/shopware-6-en/settings/Numberranges

The current implementation in this pull request is just a base for discussing this proposal.

### DIFF
--- a/src/CoreShop/Component/Core/Order/NumberGenerator/SaleNumberGenerator.php
+++ b/src/CoreShop/Component/Core/Order/NumberGenerator/SaleNumberGenerator.php
@@ -72,7 +72,18 @@ final class SaleNumberGenerator implements NumberGeneratorInterface
         }
 
         if ($store instanceof StoreInterface) {
-            return sprintf('%s%s%s', $this->configurationService->getForStore($this->prefixConfigurationKey, $store), $this->numberGenerator->generate($model), $this->configurationService->getForStore($this->suffixConfigurationKey, $store));
+            $prefix = $this->configurationService->getForStore($this->prefixConfigurationKey, $store);
+            $suffix = $this->configurationService->getForStore($this->suffixConfigurationKey, $store);
+            
+            //Implement some useful replacements
+            $prefix = str_replace('{date}',date('Ymd'),$prefix);
+            $prefix = str_replace('{year}',date('Y'),$prefix);
+            $prefix = str_replace('{month}',date('m'),$prefix);
+            $prefix = str_replace('{day}',date('d'),$prefix);
+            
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
Still not dynamic enough. Can you change it to use at least the expression engine?

Cruiser13:
@dpfaffenbauer I agree, this is just a base for discussing this feature. We should consider more options than just using the few date entries I put in there. Depending on the complexity we would like to include we should go for different solutions.

dpfaffenbauer:
Expression Engine... that is extendable

Cruiser13:
Changed accordingly. Not sure if we should add other contexts though (as coreshop does with the ExpressionInterpreter).

dpfaffenbauer:
... please use proper dependency injection

src/CoreShop/Component/Core/Order/NumberGenerator/SaleNumberGenerator.php
```diff
@@ -72,7 +73,14 @@ public function generate(ResourceInterface $model)
         }
 
         if ($store instanceof StoreInterface) {
-            return sprintf('%s%s%s', $this->configurationService->getForStore($this->prefixConfigurationKey, $store), $this->numberGenerator->generate($model), $this->configurationService->getForStore($this->suffixConfigurationKey, $store));
+            $prefix = $this->configurationService->getForStore($this->prefixConfigurationKey, $store);
+            $suffix = $this->configurationService->getForStore($this->suffixConfigurationKey, $store);
+            
+            $expr = new ExpressionLanguage();
```
dpfaffenbauer: inject via DI

Cruiser13:
Sorry @dpfaffenbauer I missed your comments. Thanks for the modification!

dpfaffenbauer:
Actually, thinking about it, this is a BC break.

Since currently we have default settings like prefix "O" for Order. But, if you pass "O" into the Expression Engine, it throws an exception, cause "O" is not a valid expression, should be "\"O\"" to be valid.

If you need that feature, you'd have to wait for 3.0, or decorate the Generator in your project and add that

Cruiser13:
@dpfaffenbauer as 3.0 is now prepared this might be a good include for 3.0?

dpfaffenbauer:
Yes.

PR #1210, opened 2020-01-06
ID Argument for re-index command

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

This PR adds an argument to the `coreshop:index` command, which allows defining certain index IDs to re-index. Default stays the same: all indices get re-indexed.

Additionally, if no indices are found (either by ID lookup or `findAll` method) a message is displayed in the console.


dpfaffenbauer:
Don't use ID's, use names. WDYT?

aarongerig:
Names would be great as well, what if we make both possible (IDs and names)?

dpfaffenbauer:
sure, ok

aarongerig:
Done

dpfaffenbauer:
ok, can you now base the PR to the 2.1 branch?

Issue #1209, opened 2020-01-06
[FR] Maximum Quantity to Order


HuppertzL:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | yes

I'm currently working on a Shop which is also selling Tickets for a Venue.
Currently Stockkeeping of a Product is able to handle a "Minimum Quantity To Order".
But for cases like Tickets it would also be good to have a "Maximum Quantity To Order" to be able to set a maximum Limit to avoid bulk purchases of some black market dealers.

dpfaffenbauer:
Good idea, would you be able to provide a PR with it?

HuppertzL:
I'll try, but i'm not familiar with the Behat thingy.
So if it's okay, i'd just focus on the business logic itself.

PR #1217, opened 2020-01-06
Feature/gh 1209 add maximum quantity to order

HuppertzL:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1209    

Added "maximumQuantityToOrder" to DataObject-Definition (json) and all related Models + Interfaces (Product + ProductInterface).
Also extended the Validators handling the maximumQuantity (AddToCartQuantityValidator, CartQuantityValidator)
Last but not least added the translations for the Constraint message keys.

But didn't implement anything of Behat.


dpfaffenbauer:
I would have added a separate Constraint and Validator to handle that, what do you think about that idea?

Also, can you add the migration for the new "maximumToOrder" field?

HuppertzL:
Well, i used the same Constraint due to  to its name, because maximum and minimum is both a quantity. But of course i could separate them if necessary.

I have to check that, i'm not that familiar with migrations.

dpfaffenbauer:
I'd deprecate the QuantityConstraint and make a MinQuantity and MaxQuantity Constraint instead. Just to keep it cleaner and more separate. I could help with that if you don't know how to deprecate stuff

HuppertzL:
Hm. Forgot to do deprecations. I'll add them

src/CoreShop/Bundle/CoreBundle/Migrations/Version20200107065213.php
```diff
@@ -0,0 +1,76 @@
...
+
+    }
+
+    /**
+     * @param Schema $schema
+     * @throws \CoreShop\Component\Pimcore\Exception\ClassDefinitionNotFoundException
+     */
+    public function down(Schema $schema)
+    {
+        $productClass = $this->container->getParameter('coreshop.model.product.pimcore_class_name');
```
dpfaffenbauer: not necessary, we never make down migrations for data-objects due to potential data-loss

src/CoreShop/Bundle/CoreBundle/Validator/Constraints/CartMaximumQuantityValidator.php
```diff
@@ -0,0 +1,131 @@
...
+            $productsChecked[] = $product->getId();
+
+            if ($higherThenMaximum === true) {
+                $invalidProduct = $product;
+
+                break;
+            }
+        }
+
+        if ($higherThenMaximum === true && $invalidProduct instanceof StockableInterface) {
```
dpfaffenbauer: $higherThanMaximum not needed, since $invalidProduct is set when invalid anyway

dpfaffenbauer:
Deprecation with backward compatibility is very important

src/CoreShop/Bundle/CoreBundle/Migrations/Version20200107065213.php
HuppertzL: okay, i'll remove it

HuppertzL:
I hope i covered everything

dpfaffenbauer:
I am still not quite happy with the deprecation. What if someone uses the old Constraint/Validator? That wouldn't work now, would it? We should extend the old one from the new one and trigger a deprecation notice, shouldn't we?

HuppertzL:
I don't really know how to solve this:
With logic in the old Validator it causes a double-validation of minimumQuantity.
Because MinimumQuantityValidator and QuantityValidator are executed.

dpfaffenbauer:
I'ved fixed it. The idea is to still have the AddToCartQuantity Constraint and Validator as well as the CartQuantity constraint and CartQuantityValidator available in the source code. 

It still is true that it might happen that the validation is triggered twice if you have a custom validation file for the cart. But we can mitigate that with a change log entry.

dpfaffenbauer:
now we have to rebase to the 2.1 branch, can you do that?

HuppertzL:
I don't really understand why Travis / phpstan is failing.
```
 ------ ---------------------------------------------------------------------------------------- 
  Line   CoreShop/Bundle/CoreBundle/Validator/Constraints/AddToCartMinimumQuantityValidator.php  
 ------ ---------------------------------------------------------------------------------------- 
  100    Access to an undefined property                                                         
         Symfony\Component\Validator\Constraint::$messageBelowMinimum.                           
 ------ ---------------------------------------------------------------------------------------- 
```

I think it's not recognizing that the Constraint is already checked for the right class by `Assert`.
But i don't understand why this is failing only once because this snippet is occurring on several places. Is it because of the `set -e` that it's aborting on the first occurrence?

I don't know how to fix that except using `instanceof` which would be duplicate to the `Asserts`, i think.
`@var AddToQuantity $constraint` doesn't work.

Do you have a suggestion?

HuppertzL:
> now we have to rebase to the 2.1 branch, can you do that?

did you mean to change the base of the PR or a git rebase?
atm i did the first option but could do the second either

dpfaffenbauer:
you also need to rebase your changes to the 2.1 branch, otherwise we also have all the commits from master here...

HuppertzL:
i don't think that this worked properly.
it was duplicating the commits. commits we wanted to get rid of

HuppertzL:
@dpfaffenbauer 
I messed this one up i think and therefore created #1222

PR #1241, opened 2020-01-20
fix: $indexIds is always an array, hence the condition is now empty n…

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

In my previous PR I added the argument `$indexIds` to the re-index command. Since `$indexIds` always returns an array (not null) even if not defined in the command line, the first condition was always false. By checking for `empty()` we make sure the condition is applied correctly.

Sorry for the mistake!


dpfaffenbauer:
thanks

PR #1252, opened 2020-01-28
[CoreBundle] Change repository so that unit definition deletion works with multiple product models 

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

The validation controller for product unit definitions only worked for the default product model and not with multiple. Now it can query all product models via the stack repository.


dpfaffenbauer:
thanks

PR #1267, opened 2020-02-07
[Cart] introduce cart-context resolver to allow better extendability of the context used for the cart

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no

### DIFF
--- a/src/CoreShop/Component/Core/Cart/CartContextResolver.php
+++ b/src/CoreShop/Component/Core/Cart/CartContextResolver.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * CoreShop.
+ *
+ * This source file is subject to the GNU General Public License version 3 (GPLv3)
+ * For the full copyright and license information, please view the LICENSE.md and gpl-3.0.txt
+ * files that are distributed with this source code.
+ *
+ * @copyright  Copyright (c) 2015-2020 Dominik Pfaffenbauer (https://www.pfaffenbauer.at)
+ * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
+ */
+
+namespace CoreShop\Component\Core\Cart;
+
+use CoreShop\Component\Cart\Cart\CartContextResolverInterface;
+use CoreShop\Component\Core\Model\StoreInterface;
+use CoreShop\Component\Order\Model\CartInterface;
+use Webmozart\Assert\Assert;
+
+class CartContextResolver implements CartContextResolverInterface
+{
+    /**
+     * @var CartContextResolverInterface
+     */
+    private $inner;
+
+    /**
+     * @param CartContextResolverInterface $inner
+     */
+    public function __construct(CartContextResolverInterface $inner)
+    {
+        $this->inner = $inner;
+    }
+
+    public function resolveCartContext(CartInterface $cart)
+    {
+        $context = $this->inner->resolveCartContext($cart);
+
+        $store = $cart->getStore();
+
+        /**
+         * @var StoreInterface $store
+         */
+        Assert::isInstanceOf($store, StoreInterface::class);
+
+        $context['store'] = $store;
+        $context['country'] = $store->getBaseCountry();
+
+        return $context;
+    }
+}
--- a/src/CoreShop/Component/Core/Cart/CartContextResolver.php
+++ b/src/CoreShop/Component/Core/Cart/CartContextResolver.php
@@ -17,7 +17,7 @@ use CoreShop\Component\Core\Model\StoreInterface;
 use CoreShop\Component\Order\Model\CartInterface;
 use Webmozart\Assert\Assert;
 
-class CartContextResolver implements CartContextResolverInterface
+final class CartContextResolver implements CartContextResolverInterface
 {
     /**
      * @var CartContextResolverInterface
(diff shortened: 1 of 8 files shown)


src/CoreShop/Component/Core/Order/Processor/CartItemsProcessor.php
```diff
@@ -46,36 +52,48 @@ final class CartItemsProcessor implements CartProcessorInterface
...
 
     /**
      * {@inheritdoc}
      */
     public function process(CartInterface $cart)
     {
-        $store = $cart->getStore();
+        if (null === $this->cartContextResolver) {
+            @trigger_error(
+                'Using CartItemsProcessor without a CarContextResolverInterpreter is deprecated since 2.1.2 and will be removed with 3.0.0',
```
jdreesen:
```suggestion
                'Using CartItemsProcessor without a CartContextResolverInterface is deprecated since 2.1.2 and will be removed with 3.0.0',
```

src/CoreShop/Component/Core/Order/Processor/CartItemsWholesaleProcessor.php
```diff
@@ -22,38 +23,56 @@
...
 
     /**
      * {@inheritdoc}
      */
     public function process(CartInterface $cart)
     {
-        $store = $cart->getStore();
+        if (null === $this->cartContextResolver) {
+            @trigger_error(
+                'Using CartItemsWholesaleProcessor without a CarContextResolverInterpreter is deprecated since 2.1.2 and will be removed with 3.0.0',
```
jdreesen:
```suggestion
                'Using CartItemsWholesaleProcessor without a CartContextResolverInterface is deprecated since 2.1.2 and will be removed with 3.0.0',
```

src/CoreShop/Component/Order/Cart/CartContextResolver.php
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @copyright  Copyright (c) 2015-2020 Dominik Pfaffenbauer (https://www.pfaffenbauer.at)
+ * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
+ */
+
+namespace CoreShop\Component\Cart\Cart;
+
+use CoreShop\Component\Order\Model\CartInterface;
+
+class CartContextResolver implements CartContextResolverInterface
```
jdreesen: Maybe this class should be `final`?

src/CoreShop/Component/Core/Cart/CartContextResolver.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace CoreShop\Component\Core\Cart;
+
+use CoreShop\Component\Cart\Cart\CartContextResolverInterface;
+use CoreShop\Component\Core\Model\StoreInterface;
+use CoreShop\Component\Order\Model\CartInterface;
+use Webmozart\Assert\Assert;
+
+class CartContextResolver implements CartContextResolverInterface
```
jdreesen: Maybe this class should be `final`?

src/CoreShop/Component/Order/Cart/CartContextResolverInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
+ */
+
+namespace CoreShop\Component\Cart\Cart;
+
+use CoreShop\Component\Order\Model\CartInterface;
+
+interface CartContextResolverInterface
+{
+    public function resolveCartContext(CartInterface $cart);
```
jdreesen: Other methods that use the context this method returns expect it to be an `array`, so the signature should be:

```suggestion
    public function resolveCartContext(CartInterface $cart): array;
```

dpfaffenbauer: True, but we currently don't have return-types at all in CoreShop. That will change with CoreShop 3.0. Until that, we don't introduce new ones neither.

src/CoreShop/Component/Core/Cart/CartContextResolver.php
dpfaffenbauer: done

src/CoreShop/Component/Order/Cart/CartContextResolver.php
dpfaffenbauer: done

src/CoreShop/Component/Core/Order/Processor/CartItemsWholesaleProcessor.php
dpfaffenbauer: done

src/CoreShop/Component/Core/Order/Processor/CartItemsProcessor.php
dpfaffenbauer: done

src/CoreShop/Component/Order/Cart/CartContextResolverInterface.php
jdreesen: Oh, okay... in that case there should be some PhpDoc at least, I think.

```suggestion
    /**
     * @return array
     */
    public function resolveCartContext(CartInterface $cart);
```

dpfaffenbauer: done

PR #1283, opened 2020-02-14
feature: calculate shipping tax using cart items

lukadschaak:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Moin!
@dpfaffenbauer , do you remember the shipping tax calculation in germany? We spoke about that in the training and you implemented some stuff.

This PR implements the possibility to select a service, that will calculate the tax amount of the shipping price. In some countries the shipping price will have fix tax rate. In other countries, for example in Germany, the tax amount for the shipping cost is variable. It depends on the tax rate of the cart items. When all Items have the same tax rate, the shipping tax is the same (7% or 19%). If there are items with different tax rates in cart, it gets complicated.

With this Feature you can implement your own service and define your own tax calculation. Also the two ways of calculation are pre-implemented.

The Code is WIP. First of all i want to know if @dpfaffenbauer wants this feature in core and if you are happy with this way of implementing this feature. If yes i would finish the code and try to add tests.

On the following screen you see the combobox i added.

cart-based-taxation


dpfaffenbauer:
I'd much rather go a step further and would introduce shipping calculation strategies, where you can register services, these services gets displayed in a combobox and you can choose between them, wdyt?

lukadschaak:
Like we start with our two strategies, and register them for example as service with the symfony tagging system? Then people are able to register their own service, which will then be displayed in the combobox?!
I think that sounds really nice. So every Country can select/implement his fitting service.

dpfaffenbauer:
yep

dpfaffenbauer:
BTW: If you are not sure how to implement it, we have several examples of how that is done in CoreShop. eg. CoreShop Index

lukadschaak:
I will take a look at the current implementations and update this PR asap.

lukadschaak:
What is the reason of adding the fields 'Stores' and 'Tax Rate' (see Screenshot of PR description) in ExtJS via `panel.down("fieldset").add` in a seperate item.js inside the CoreBundle?

`src/CoreShop/Bundle/CoreBundle/Resources/public/pimcore/js/shipping/carrier/item.js`

dpfaffenbauer:
cause the shipping bundle doesn't know anything about stores nor taxes. only the core-bundle does

src/CoreShop/Bundle/CoreBundle/Migrations/Version20200213132916.php
```diff
@@ -0,0 +1,28 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20200213132916 extends AbstractPimcoreMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE coreshop_carrier ADD taxStrategy VARCHAR(255) DEFAULT NULL AFTER logo;');
```
dpfaffenbauer: shouldn't we migrate existing carriers to tax-rule?

src/CoreShop/Bundle/ShippingBundle/DependencyInjection/Compiler/ShippingTaxStrategyPass.php
```diff
@@ -0,0 +1,30 @@
...
+ * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
+ */
+
+namespace CoreShop\Bundle\ShippingBundle\DependencyInjection\Compiler;
+
+use CoreShop\Bundle\PimcoreBundle\DependencyInjection\Compiler\RegisterRegistryTypePass;
+
+final class ShippingTaxStrategyPass extends RegisterRegistryTypePass
+{
+    public const SHIPPING_TAX_STRATEGY_TAG = 'coreshop.shipping.tax.strategy';
```
dpfaffenbauer: Please also provide for autoloading, see Extension for more.

src/CoreShop/Bundle/ShippingBundle/Resources/config/serializer/Model.Carrier.yml
```diff
@@ -23,11 +23,18 @@ CoreShop\Component\Shipping\Model\Carrier:
             expose: true
             type: pimcoreAsset
             groups: [Detailed]
+        taxStrategy:
+            expose: true
+            type: string
+            groups: [Detailed]
+        cartBasedTaxation:
```
dpfaffenbauer: remove `cartBasedTaxation`

src/CoreShop/Bundle/ShippingBundle/Resources/config/services/shipping-rules.yml
```diff
@@ -24,6 +24,17 @@ services:
     coreshop.form_registry.shipping_rule.actions:
         class: CoreShop\Bundle\ResourceBundle\Form\Registry\FormTypeRegistry
 
+    coreshop.registry.shipping.tax.strategies:
```
dpfaffenbauer: move to `shipping-taxation.yml `

src/CoreShop/Bundle/ShippingBundle/Resources/public/pimcore/js/carrier/item.js
```diff
@@ -125,6 +125,16 @@ coreshop.carrier.item = Class.create(coreshop.resource.item, {
                             bodyStyle: 'padding:10px;'
                         },
                         items: langTabs
+                    },
+                    {
+                        xtype: 'combo',
+                        name: 'taxStrategy',
+                        fieldLabel: 'Steuerberechnung',
```
dpfaffenbauer: use translation

src/CoreShop/Bundle/ShippingBundle/Resources/translations/admin.de.yml
```diff
@@ -30,4 +29,6 @@ coreshop_condition_dimension: 'Dimension'
 coreshop_action_additionAmount: 'Aufschlag Menge'
 coreshop_action_additionPercent: 'Aufschlag Prozent'
 coreshop_shipping: 'Versand'
-coreshop_condition_amount_gross: 'Nutze Bruttowert'
+coreshop_shipping_tax_strategy: 'Steuerberechnung'
```
dpfaffenbauer: please don't provide german translations, they'll get overwritten by crowdin anway

src/CoreShop/Component/Core/Order/Processor/CartTaxProcessor.php
```diff
@@ -40,19 +39,28 @@ final class CartTaxProcessor implements CartProcessorInterface
...
+    /**
+     * @var ServiceRegistry
+     */
+    private $registry;
+
     /**
      * @param TaxCollectorInterface         $taxCollector
      * @param AddressProviderInterface      $defaultAddressProvider
      * @param TaxCalculatorFactoryInterface $taxCalculatorFactory
+    //     * @param iterable                      $taxStrategyTaggedServices
```
dpfaffenbauer: wrong syntax

tests/lib/CoreShop/Test/Data.php
```diff
@@ -53,6 +53,11 @@ class Data
      */
     public static $product3;
 
+    /**
```
dpfaffenbauer: please don't use PHPUnit for these kind of tests. Most of them get removed with 3.0 as we primarily test with behat.

src/CoreShop/Bundle/ShippingBundle/Resources/translations/admin.de.yml
lukadschaak: shall i just delete the german translations?

dpfaffenbauer: yes pls

src/CoreShop/Bundle/CoreBundle/Migrations/Version20200213132916.php
lukadschaak: Nice hint. I agree.

tests/lib/CoreShop/Test/Data.php
lukadschaak: I am not familiar with Behat. It looks like everything is written in Gherkin. So you will translate all tests to behat in 3.0?
In `ShippingTaxationCartItemsTest.php` i wrote a complex mocking test with prophecy and some simpler tests by using the container from Kernel. Shall i remove everything abount testing i added in this PR or only the complex prophecy test? Or did you mean something else?

lukadschaak: Ah, after thinking a bit further i understand what you mean by not using phpunit tests.
Have you an idea how to test the service? What about testing "update the cart" ?

dpfaffenbauer: Actually, all of these tests are already done withi behat: https://github.com/coreshop/CoreShop/tree/master/features

We don't do prophecy, you should be able to create the same with behat. So we actually test the behaviour of the application. If you don't know how to do that, I can help you there.

lukadschaak: I will look at it. Thanks for the moment.

dpfaffenbauer: Update the cart is already covered. Testing the service directly is not something you would do with behat. You can, but we currently don't. We are testing actually against the cart. So would say something like:


```gherkin
@cart
Feature: Use a different calculation strategy for shipping tax calculation

  Background:
    Given the site operates on a store in "Austria"
    And the site has a currency "Swiss Franc" with iso "CHF"
    And I am in country "Austria"
    And the site has a tax rate "CH" with "7.7%" rate
    And the site has a tax rule group "CH"
    And the tax rule group has a tax rule for country "Austria" with tax rate "CH"
    And the site has a carrier "Post"
    And the carrier "Post" uses tax-calculation strategy "cart-items"
    And adding a shipping rule named "post"
    And the shipping rule is active
    And the shipping rule has a action price of 10 in currency "CHF"
    And the shipping rule belongs to carrier "Post"
    And the site has a product "Shoe" priced at 10000
    And the product has the tax rule group "CH"

  Scenario: Create a new cart, add a product and shipping should be applied
    And I add the product "Shoe" to my cart
    Then the cart shipping should be "1000" excluding tax
    And the cart shipping should be "1077" including tax
    And the loaded carts shipping tax rate should be "7.7"
```

Note that the values are not right there yet, should just get you a feeling how I would dom it

lukadschaak:
General question. I'm thinking about naming and maybe ShippingTaxCaluculation would be more appropiate than ShippingTaxStrategy?!

dpfaffenbauer:
`ShippingTaxCalculationStrategy`?

lukadschaak:
Sounds good. I will do it

dpfaffenbauer:
Is this ready to be reviewed and tested?

FYI: Travis currently fails cause Pimcore 6.5.2 is broken...

src/CoreShop/Bundle/ShippingBundle/Resources/config/pimcore/routing.yml
```diff
@@ -1,7 +1,17 @@
+# https://docs.coreshop.org/2.1.0/Bundles/Resource_Bundle/CustomEntities.html
+# If you want to add additional routes
+# - Add a new Controller which extends ResourceController
+# - Change the DI Configuration: CoreShop/Bundle/ShippingBundle/DependencyInjection/Configuration.php
```
dpfaffenbauer: where does that come from?

src/CoreShop/Component/Shipping/Taxation/TaxCalculationStrategyCartItems.php
```diff
@@ -0,0 +1,131 @@
...
+        $this->updateShippingAdjustment($cart, $shippingAdjustment, $shippingTaxAmount);
+
+        return $usedTaxes;
+    }
+
+    private function collectCartItemsTaxRules(CartInterface $cart, array &$totalAmount, array &$taxRules): void
+    {
+        /** @var CartItemInterface $item */
+        foreach ($cart->getItems() as $item) {
+            if ($item->getDigitalProduct() === true) {
```
dpfaffenbauer: I think there is a function called isShippable? not sure though

lukadschaak: Cannot find any occurrence of that function

lukadschaak:
> Just realised, this cannot be finished yet. Carrier Price Calculation also knows about tax and needs to use the tax-calc service as well, right?

Not sure where else the Carrier Price Calculation takes place. I just know of the CartProcessors which listen to cart updates. But yes, The Carrier price calculation always depends on these tax-calc services i think.

dpfaffenbauer:
@solverat could you review that as well pls?

dpfaffenbauer:
big thanks @lukadschaak

PR #1303, opened 2020-02-24
[IndexBundle] allow configuring if versions should be indexed or not

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no


dpfaffenbauer:
Tests fail due to Pimcore. Nothing I can currently do... still have to merge it though.

src/CoreShop/Bundle/CoreBundle/Migrations/Version20200224160300.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace CoreShop\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Pimcore\Migrations\Migration\AbstractPimcoreMigration;
+
+class Version20200224160300 extends AbstractPimcoreMigration
```
jdreesen: Why do you need two migrations doing the same (in different bundles)?

dpfaffenbauer: cause of separation of bundles. When you install full-suite CoreShop, the IndexBundle is not installed, only the CoreBundle. If you only use the IndexBundle, you don't have the CoreBundle installed.... And since it's just migrations, I am happy with the duplication.

Issue #1307, opened 2020-02-28
Prevent direct calls of fragments


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Prevent direct calls of fragments like this: https://demo2.coreshop.org/en/shop/en/shop/cart/add/62
 
My Approach: 

```yml
# Prevent direct fragment calls
AppBundle\EventListener\FragmentProtectionListener:
    tags:
        - { name: kernel.event_listener, event: kernel.controller, method: checkFragmentRequest }
```

```php
<?php

namespace AppBundle\EventListener;

use Symfony\Component\HttpKernel\Event\ControllerEvent;
use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;

class FragmentProtectionListener
{
   // make them configurable in options?
    const PROTECTED_ROUTES = [
        'coreshop_cart_add'
    ];

    /**
     * @param ControllerEvent $event
     */
    public function checkFragmentRequest(ControllerEvent $event)
    {
        $request = $event->getRequest();

        if (!$event->isMasterRequest()) {
            return;
        }

        if(!$request->isMethod('GET')) {
            return;
        }

        if (!$request->attributes->has('_route')) {
            return;
        }

        $route = $request->attributes->get('_route');
        if (!in_array($route, self::PROTECTED_ROUTES)) {
            return;
        }

        throw new NotFoundHttpException();
    }
}
```

solverat:
instead of NotFoundHttpException we could redirect the user to the cart.

PR #1309, opened 2020-02-28
[FrontendBundle/CoreBundle] prevent _fragment calls by using ACL's

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1307 

With two different routes, one for _partial rendering, one for actual action call. We differentiate these calls by using default entries in Pimcore's Staticroutes


solverat:
Missing:
- Migration for new staticroute
- Changelog: a very, very very bold note, that everyone should update the routes in project views!

PR #1315, opened 2020-03-04
[Customer] fix missing username field

dpfaffenbauer:
otherwise Symfony fails with a not-implemented exception.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


src/CoreShop/Component/Customer/Model/Customer.php
```diff
@@ -120,22 +120,6 @@ public function setEmail($email)
...
-     */
-    public function getUsername()
-    {
-        throw new ImplementedByPimcoreException(__CLASS__, __METHOD__);
-    }
-
-    /**
-     * @param string $username
-     */
-    public function setUsername($username)
```
solverat: Keep setUsername please

src/CoreShop/Component/Customer/Model/Customer.php
```diff
@@ -216,6 +200,14 @@ public function setCustomerGroups($customerGroups)
         throw new ImplementedByPimcoreException(__CLASS__, __METHOD__);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getUsername()
+    {
+        return $this->getEmail();
```
solverat: add some notes? like "fallback, if no real username has been defined"

src/CoreShop/Bundle/CustomerBundle/Form/Type/CustomerType.php
```diff
@@ -66,7 +69,14 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'invalid_message' => 'coreshop.form.customer.email.must_match',
                 'first_options' => ['label' => 'coreshop.form.customer.email'],
                 'second_options' => ['label' => 'coreshop.form.customer.email_repeat'],
+            ])
+        ;
+
+        if ($options['allow_username']) {
+            $builder->add('username', TextType::class, [
+                'label' => 'coreshop.form.customer.username',
```
solverat: Thee username field gets appended, so the username input field will show up after the submit button.  Could we change the priority?

solverat: na - my fault. Everything is fine here. I had my own form layout in twig.

src/CoreShop/Component/Customer/Model/Customer.php
solverat: And also add it to `CoreShop\Component\Customer\Model\CustomerInterface`?

PR #1324, opened 2020-03-17
[3.0] change cart/order base-currency conversion

dpfaffenbauer:
base-currency ALWAYS is Store Currency, where as the display currency is the one we need to convert

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


UPGRADE-3.0.x.md
```diff
@@ -1,5 +1,73 @@
...
+## Cart and Quote are now Order
+The Cart and Quote Class are now the Order Class as well. Since these entities are similar, it makes sense to just use one.
+
+### Migration
+CoreShop will not provide any migration Script to convert carts/quotes into Orders. That is impossible for us to provide.
+The Order is basically the same as before, so you can create custom migrations to move them.
+
+Only thing we migrate is for existing orders eg. add new Fields or set the initial state.
+
+### Differential between Cart, Quote and Order
```
jdreesen, line 16:
```suggestion
### Difference between Cart, Quote and Order
```

dpfaffenbauer: 🤦 thanks

PR #1330, opened 2020-03-20
Password reset redirect to home if user not exists

grammit:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

If the user does not exist, an error message will appear


dpfaffenbauer:
1. Please sign the CLA
2. I don't think that the current solution is right nor yours. I think we should also redirect to home with the success message. Otherwise you could exploit that and check what Users exist and what not

dpfaffenbauer:
closed due to inactivity

PR #1340, opened 2020-04-02
[Pimcore Bundle] Fix loading dynamic dropdown options

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

This PR fixes sanitizing the parent folder path when trying to load the options for dynamic dropdown tags. It uses the same validation as Pimcore. Furthermore, the method now returns an empty array, if the folder could not be found.


dpfaffenbauer:
can you check why travis fails?

aarongerig:
Should run through now...

dpfaffenbauer:
thanks

PR #1348, opened 2020-04-14
[DynamicDropdown] add dependency resolving

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


aarongerig:
@dpfaffenbauer Just tested this on my local instance and it seems to work. Dependencies were resolved both ways (requires and required by). 👍

PR #1354, opened 2020-04-20
[3.0] introduce SEO Schema Generator

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "rinvex/countries": "^6.0",
         "stof/doctrine-extensions-bundle": "^1.2",
         "symfony/symfony": "^4.4",
-        "webmozart/assert": "^1.2"
+        "webmozart/assert": "^1.2",
+        "spatie/schema-org": "^2.13"
     },
     "require-dev": {
         "behat/behat": "^3.3",
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
         "stof/doctrine-extensions-bundle": "^1.2",
         "symfony/symfony": "^4.4",
         "webmozart/assert": "^1.2",
-        "spatie/schema-org": "^2.13"
+        "spatie/schema-org": "^2.8"
     },
     "require-dev": {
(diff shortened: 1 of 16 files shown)


composer.json
```diff
@@ -35,7 +35,8 @@
         "rinvex/countries": "^6.0",
         "stof/doctrine-extensions-bundle": "^1.2",
         "symfony/symfony": "^4.4",
-        "webmozart/assert": "^1.2"
+        "webmozart/assert": "^1.2",
+        "spatie/schema-org": "^2.13"
```
jdreesen, line 39: Careful, because this seems to need PHP `^7.3` while CoreShop is `^7.2`.

dpfaffenbauer: thanks, tests also fail cause of that.

solverat:
I'm not sure about this. This is basically the same as [dachcom-digital/pimcore-schema](https://github.com/dachcom-digital/pimcore-schema) does. We should consider removing the seo bundle and start using [dachcom-digital/pimcore-seo]/https://github.com/dachcom-digital/pimcore-seo) instead, to finally get rid of thousands of different seo implementation approaches.

This could be a good new feature for 3.0. Also: I also don't like the idea to use a request listener for all of this (Yea i know, i've done this by myself within the schema bundle...). Instead, it's way nicer to use a simple twig helper, as described [here](https://github.com/dachcom-digital/pimcore-seo/blob/master/docs/00_Setup.md#default-layout).

Join forces to build the last and final seo bundle forever is definitely something we need here, right? :) WDYT?

dpfaffenbauer:
@solverat I agree, but CoreShop' SEO Bundle does more than just schema, that's why I added this here. Might be something for friends-of-pimcore?

dpfaffenbauer:
decided against that, with 3.0 we gonna deprecate the SEO Bundle and use the dachcom bundle

PR #1399, opened 2020-05-12
[SEO] integration to dachcom-digital/pimcore-seo

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      |  no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | #...


dpfaffenbauer:
@solverat pls review

src/CoreShop/Bundle/SEOBundle/README.md
```diff
@@ -1,6 +1,8 @@
 CoreShop SEO Bundle
 ===================
 
+# Deprecated due to https://github.com/dachcom-digital/pimcore-seo, will be removed with 3.1
```
jdreesen, line 4: Wouldn't it be better to deprecate it in `2.x` and remove it in `3.0`? Not sure if it violates SemVer otherwise...

dpfaffenbauer: :+1, we changed that now anyway. we keep our SEO Bundle and suggest the usage of dachcom/seo instead. Since we don't wanna deeply couple the both

PR #1408, opened 2020-05-22
Fixed issue with saving address changes

ramundomario:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Fixed issue with saving address changes, reproducible on [demo](https://demo2.coreshop.org):
bug

Issue is caused by `CustomerTransformHelper` responsible for saving address, but saving is done only in case the address is new or parent is different.
https://github.com/coreshop/CoreShop/blob/9228f1a6d346deb934387c32c033dd5c172d791f/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php#L175-L178


dpfaffenbauer:
thanks

solverat:
@ramundomario nice catch! Thanks!

PR #1433, opened 2020-06-19
[OrderBundle] Values should be zero, if amount should not be defined

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

There is a check if a value is 0 in the condition checker, but 0s are not valid in the given form configuration


dpfaffenbauer:
thanks

solverat:
Also port to 2.2.3

Issue #1257, opened 2020-02-04
"Invoice Address is Shipping Address" checkbox results in form error


GALCF:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Coreshop version 2.2.0 (current demo)
Pimcore version 6.4.2

While ordering with an invoice address that is not a shipping address, checking "Invoice Address is Shipping Address" and submitting the form results in a form error that is not shown because the checkbox hides the error with the shipping address input.

**Expanded View (error not visible while checked):**

![image](https://user-images.githubusercontent.com/14803394/73734031-48e89280-473d-11ea-8f4c-e1405b4f926f.png)

dpfaffenbauer:
ok, so that is only a UI problem right?

GALCF:
Yes, UI as in the error cannot be seen, or maybe a "workflow" problem, that some invoice addresses may not be a shipping address too (if the checkbox is present, which is also UI-only?)

dpfaffenbauer:
Yes. I'd say thats just how you display things. Since the introduction of address-identifiers that problem can happen. Not sure how to solve properly.... Guess we should have a different select for addresses allowed for shipping as well and swap these two based on the checkbox.

solverat:
@GALCF could you please check the user: just open the cart customer, then open the address object from "default address". Is there a defined "address identifier"?
 
![image](https://user-images.githubusercontent.com/700119/86234251-ecc45600-bb96-11ea-8665-80544e6a9900.png)

solverat:
@GALCF btw, we're using this in frontend: 

```twig
{% set hide_shipment_selector = hasShippableItems == false or (form.useInvoiceAsShipping is defined and form.useInvoiceAsShipping.vars.data and form.shippingAddress.vars.valid != false) %}
```
 instead of

```twig 
{% if hasShippableItems == true and form.useInvoiceAsShipping.vars.data %}
```

which should be the default i think.

solverat:
I'll fix this in my latest PR (https://github.com/coreshop/CoreShop/pull/1440).

dpfaffenbauer:
fixed

PR #1440, opened 2020-07-01
assert default address type

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | /no
| Fixed tickets | #1257

This PR will:
- assert if a customers default address is valid within a given address identifier (We only have one default address, but multiple types of address identifier). 
- check if valid shipping addresses are available in address checkout step. if not, disable `useInvoiceAsShipping` checkbox since it is not possible to use it
- add validation groups to the `NotBlank` constraints
- Improve UI in address checkout step: if `shippingAddress` is not valid, show selector after page reloads.
- If `useInvoiceAsShipping` has been checked, check if invoiceAddress can be assigned to shippingAddress first - if not, throw a proper exception. To be clear: This is just a development improvement to speed up the issue tracking time. The check itself should be done after a user selects an address (by JS for example).


solverat:
Also wrong, needs to be checked if invoiceAddress can be assigned to shippingAddress first!
 
https://github.com/coreshop/CoreShop/blob/c429578588c7652d255728d260a2f2eaf34d38ed/src/CoreShop/Bundle/CoreBundle/Form/Type/Checkout/AddressType.php#L101

Issue #1448, opened 2020-07-10
[Voucher] Voucher code generator returns always the same code


albertmueller:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

PHP 7.4
Pimcore 6.6.9
CoreShop dev-master

There is a issue in the generateCode function, i got always the same code because of the int type declaration.

PR #1449, opened 2020-07-10
fix voucher code generator

albertmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1448

### DIFF
--- a/src/CoreShop/Component/Order/Generator/CartPriceRuleVoucherCodeGenerator.php
+++ b/src/CoreShop/Component/Order/Generator/CartPriceRuleVoucherCodeGenerator.php
@@ -77,7 +77,11 @@ class CartPriceRuleVoucherCodeGenerator
 
     protected static function generateCode(string $letters, int $length): string
     {
-        mt_srand(intval(microtime() * 1000000));
+        $rand = (int) microtime()*1000000;
+        if ($rand === 0) {
+            $rand = intval(microtime()*1000000);
+        }
+        mt_srand($rand);
         $i = 0;
         $code = '';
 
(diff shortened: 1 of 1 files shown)


src/CoreShop/Component/Order/Generator/CartPriceRuleVoucherCodeGenerator.php
```diff
@@ -77,17 +77,30 @@ public function generateCodes(CartPriceRuleVoucherGeneratorInterface $generator)
...
-            $num = mt_rand() % (strlen($letters));
-            $tmp = substr($letters, $num, 1);
-            $code = $code . $tmp;
-            $i++;
+        $code = "";
+        $max = strlen($letters);
+
+        for ($i=0; $i < $length; $i++) {
+            $letter = self::cryptoRandSecure(0, $max-1);
+            $code .= $letters[$letter];
```
jdreesen, line 85: As we're on PHP 7 here, this can/should be simplified to use the PHP native [`random_int()`](https://www.php.net/manual/de/function.random-int.php) function instead of implementing a custom one (see the update of the [SO answer](https://stackoverflow.com/a/13733588) at the bottom):

```suggestion
            $code .= $letters[random_int(0, $max-1)];
    }
```

src/CoreShop/Component/Order/Generator/CartPriceRuleVoucherCodeGenerator.php
```diff
@@ -77,17 +77,30 @@ public function generateCodes(CartPriceRuleVoucherGeneratorInterface $generator)
...
+        $bytes = (int) ($log / 8) + 1;
+        $bits = (int) $log + 1;
+        $filter = (int) (1 << $bits) - 1;
+        do {
+            $rnd = hexdec(bin2hex(openssl_random_pseudo_bytes($bytes)));
+            $rnd = $rnd & $filter;
+        } while ($rnd > $range);
+        return $min + $rnd;
+    }
+
```
jdreesen, line 105: This is not needed anymore:

```suggestion
```

albertmueller:
@jdreesen thanks, you're right. that is easier and also works :)

Issue #1463, opened 2020-07-21
Exception on sites with no store


roynilsson:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

I have a multi site setup where some of the sites are not connected to a Coreshop store. Am using latest commit (fd87c6f99aa1af520190b52793866afa2fe59c40) of the master branch.

When trying to access the site without store I get the following error.

> Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Return value of CoreShop\Component\Store\Context\RequestBased\SiteBasedRequestResolver::findStore() must implement interface CoreShop\Component\Store\Model\StoreInterface, null returned in /var/www/html/vendor/coreshop/core-shop/src/CoreShop/Component/Store/Context/RequestBased/SiteBasedRequestResolver.php on line 38

The type hint of the findStore method does not allow null values but in this case it tries to return null. I changed the type hint to allow null and this seemed to work for me. Got past the error. I am unsure if this is a good solution or not. There is a StoreNotFoundException thrown when the default store is not found. Maybe throwing that exception would be the expected behaviour? Then again would that mean that it is not possible to have sites without a store?

I think someone who knows Coreshop better than me need to figure this out.

dpfaffenbauer:
Yep, that is the right solution. Could you create a PR?

PR #1464, opened 2020-07-21
Allow null return value in request resolver find store

roynilsson:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #1463 

Changed type hint of the findStore method to allow null values. This method returns null for sites that are not connected to any store. 

BC breaks because we change the return type of a method in RequestResolverInterface.

### DIFF
--- a/src/CoreShop/Component/Store/Context/RequestBased/CompositeRequestResolver.php
+++ b/src/CoreShop/Component/Store/Context/RequestBased/CompositeRequestResolver.php
@@ -36,7 +36,7 @@ final class CompositeRequestResolver implements RequestResolverInterface
     /**
      * {@inheritdoc}
      */
-    public function findStore(Request $request): StoreInterface
+    public function findStore(Request $request): ?StoreInterface
     {
         foreach ($this->requestResolvers as $requestResolver) {
             try {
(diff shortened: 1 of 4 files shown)


dpfaffenbauer:
Not a bc break, since 3.0 (dev-master) has not been released yet.

Stan checks don't fail cause of your PR, I have to check them

jdreesen:
Would not have been a BC break anyways, because return types are covariant, which means narrowing the return type in the implementation is in compliance with LSP.

roynilsson:
But we're changing the definition of the interface. Any existing implementation would break because of the change in the interface. If this had been released already I think it should have been considered a BC break

jdreesen:
No, nothing would break. 

It's totally fine to have an interface that says
```php
interface Foo {
	function test(): ?object;
}
```
and implementations whose return types are not nullable:
```php
class Bar implements Foo {
	function test(): object {
		//...
	}
}
```

So if you add the nullability in a later release, the implementations don't need to change.

For an example, see: https://3v4l.org/KpdBO

dpfaffenbauer:
thanks

PR #1495, opened 2020-10-06
fix: ignore CartNotFoundException during interactiveLogin event

youwe-petervanderwal:
The CartBlamerListener shouldn't throw exceptions during interactive login

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

The CartBlamerListener::blame() method already have an early return on null, let the getCart() then indeed return null when there is no cart found.

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/EventListener/CartBlamerListener.php
+++ b/src/CoreShop/Bundle/CoreBundle/EventListener/CartBlamerListener.php
@@ -17,6 +17,7 @@ namespace CoreShop\Bundle\CoreBundle\EventListener;
 use CoreShop\Bundle\CoreBundle\Event\CustomerRegistrationEvent;
 use CoreShop\Component\Core\Model\CustomerInterface;
 use CoreShop\Component\Order\Context\CartContextInterface;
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
what's your use case for this? do you have logins where not cart can be available? The usual behaviour is that the cart is created if none is found.

youwe-petervanderwal:
> what's your use case for this? do you have logins where not cart can be available? The usual behaviour is that the cart is created if none is found.

We are developing a portal where the webshop only is a (relative small) part of it. When users (who had only access to the sections that aren't shop related) logged in, we ran into that exception.

Since there is a null check on line 66 on the result of getCart() it seemed that that method should return null when there isn't a cart at that moment. So either getCart() should return null on that case, or that null check is obsolete.

Anyhow, my colleagues decided to move away from CoreShop and implement the parts we need with the e-commerce bundle, so this behaviour isn't a blocker for us anymore....

dpfaffenbauer:
the use case is still valid though, created a new PR for 2.2 and I'll merge that to master later: https://github.com/coreshop/CoreShop/pull/1500

PR #1500, opened 2020-10-15
[CoreBundle] interactive login: cart might not be available for several reasons, ignore exception and don't assign a cart

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


src/CoreShop/Bundle/CoreBundle/EventListener/CartBlamerListener.php
```diff
@@ -116,6 +117,11 @@ private function blame(CustomerInterface $user)
      */
     private function getCart()
     {
-        return $this->cartContext->getCart();
+        try {
+            return $this->cartContext->getCart();
+        }
+        catch (CartNotFoundException $ex) {
+            return null;
```
jdreesen, line 124: The `@return` PhpDoc should be adapted to `@return CartInterface|null`

dpfaffenbauer: yep, thats what stan is telling as well :) Thanks though :)

PR #1503, opened 2020-10-21
Fix checkout with addressAccessType COMPANY_ONLY

geraldaistleitner:
When a customer only has access to company addresses, checkout is not possible as those addresses not show up.
This can be fixed by using the already implemented CustomerAddressAllocator also for FormType AddressChoiceType

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This pull request fixes a problem when a customer wants to do a checkout but only has access to the company addresses. The addressAccessType was not considered till now in CoreShop\Bundle\CoreBundle\Form\Type\AddressChoiceType.


dpfaffenbauer:
@geraldaistleitner Can you please rebase to 2.2?

Issue #1498, opened 2020-10-08
Document saving failed


albertmueller:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes/no

If i try to save a document, sometimes i got this error:
![image](https://user-images.githubusercontent.com/2986829/95431841-2357fc80-094e-11eb-90ef-fb76296a926e.png)

It comes from:
```
abstract class AbstractSluggableLinkGenerator implements LinkGeneratorInterface
{
    protected function slugify($string): string
    {
        return strtolower(
            trim(
                preg_replace('~[^0-9a-z]+~i', '-', html_entity_decode(
                    preg_replace(
                        '~&([a-z]{1,2})(?:acute|cedil|circ|grave|lig|orn|ring|slash|th|tilde|uml);~i',
                        '$1',
                        htmlentities($string, ENT_QUOTES, 'UTF-8')
                    ),
                    ENT_QUOTES,
                    'UTF-8'
                )),
                '-'
            )
        );
    }
}
```

alex-jungwirth:
Hi. Same error here. When returning a random string from slugify method, the stuff works. Don't know why, but could you please fix this? Pretty please <3

dpfaffenbauer:
Guess you use link-generators for objects that don't have a name?

albertmueller:
that's happen also on documents, without use of objects. "fixed" it local with a return of a empty string if $string is null

dpfaffenbauer:
It happens with documents cause of the validation check where Pimcore renders the documents thus triggering the link generator

albertmueller:
but can we return a empty string if $string is null? otherwise it is a bit confusing if you don't know whats going on. if you agree i can do a PR?

dpfaffenbauer:
yes, I guess that would be a solution

dpfaffenbauer:
that has been fixed

PR #1518, opened 2020-11-04
fix return empty string if $string is null

albertmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1498


dpfaffenbauer:
thanks

PR #1526, opened 2020-11-11
Fix checkout with addressAccessType COMPANY_ONLY

geraldaistleitner:
When a customer only has access to company addresses, checkout is not possible as those addresses not show up.
This can be fixed by using the already implemented CustomerAddressAllocator also for FormType AddressChoiceType

Q | A
-- | --
Bug fix? | yes
New feature? | no
BC breaks? | no
Deprecations? | no
Fixed tickets |  

This pull request fixes a problem when a customer wants to do a checkout but only has access to the company addresses. The addressAccessType was not considered till now in CoreShop\Bundle\CoreBundle\Form\Type\AddressChoiceType.

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Form/Type/AddressChoiceType.php
+++ b/src/CoreShop/Bundle/CoreBundle/Form/Type/AddressChoiceType.php
@@ -13,7 +13,7 @@
 namespace CoreShop\Bundle\CoreBundle\Form\Type;
 
 use CoreShop\Component\Address\Model\AddressInterface;
-use CoreShop\Component\Core\Customer\Allocator\CustomerAddressAllocator;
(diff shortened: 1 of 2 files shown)


geraldaistleitner:
Would be great if we could merge this soon 👍

dpfaffenbauer:
had to change your PR to use proper DI

dpfaffenbauer:
@solverat can we have your opinion please?

solverat:
@geraldaistleitner  sorry for my delayed response. I just checkout out your PR and of course you're right. We need to determinate the right address context of the given customer first before filtering them! 

Thanks for your investigation!

solverat (approved):

dpfaffenbauer:
thanks

PR #1562, opened 2021-01-21
[Shipping] make shipping calculation more independent from cart

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |  o

### DIFF
--- a/src/CoreShop/Component/Shipping/Model/ShippableItemInterface.php
+++ b/src/CoreShop/Component/Shipping/Model/ShippableItemInterface.php
@@ -15,6 +15,18 @@ namespace CoreShop\Component\Shipping\Model;
 interface ShippableItemInterface
 {
     /**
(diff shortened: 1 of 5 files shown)


src/CoreShop/Component/Shipping/Model/ShippableItemInterface.php
```diff
@@ -14,6 +14,18 @@
 
 interface ShippableItemInterface
 {
+    /**
+     * @param bool $withTax
+     * @return int
+     */
+    public function getTotal(bool $withTax = true);
```
dpfaffenbauer, line 21: Is actually a BC break. In Core currency only implemented from `CartItemInterface` so there it is available. Anyone against this break?

PR #1575, opened 2021-01-27
Don't submit unnamed form without any data on non-POST request.

das-peter:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | Don't know
| Deprecations? | no

I've no idea why the METHOD check of the original Smyonfy code should be removed and there's not documentation I've found which would explain why that is done or how it is supposed to work with other components.
I just got the bundle as a dependency of the dpfaffenbauer/process-manager package I installed. And now this dependency is interfering with _all_ our forms.
Most obvious issue so far is that due the permanent form submits the CSRF Token validation is triggered even on actual non-submissions, leading to bogus CSRF errors on our forms without user interaction.

Now given there's a name check for named forms in order to determine if a submission is suitable I think the least thing to do for unnamed forms is to check whether there's any data at all before just submitting empty data.
An extended approach would be to check if there are matching parameters for the form fields before triggering a submit.
Maybe something like <code>!empty(array_intersect_key($data, $form->all()))</code> as check could do - however, this would be more restrictive than the original Symfony handler and hence could lead to more trouble.

### DIFF
--- a/src/CoreShop/Bundle/ResourceBundle/Form/Extension/HttpFoundation/HttpFoundationRequestHandler.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Form/Extension/HttpFoundation/HttpFoundationRequestHandler.php
@@ -62,6 +62,10 @@ final class HttpFoundationRequestHandler extends \Symfony\Component\Form\Extensi
         if ('GET' === $method || 'HEAD' === $method || 'TRACE' === $method) {
             if ('' === $name) {
                 $data = $request->query->all();
+                // No data - nothing to submit.
+                if (!count($data)) {
+                    return;
+                }
             } else {
                 // Don't submit GET requests if the form's name does not exist
                 // in the request


dpfaffenbauer:
I don't think that change works for us. I also don't see what your problem with this might be. I use quite a lot forms with CoreShop and never ran into any issue with that. How do you use forms in Pimcore/Symfony?

das-peter:
Thanks for the swift response.
> I don't think that change works for us.

Obviously I can't comment on that as I've no clue why that adjustment is necessary in the first place :)

> I use quite a lot forms with CoreShop and never ran into any issue with that

Well, is there another component in that form handling that somehow integrates with / expects the adjusted FormRequestHandler?
If so it certainly makes sense that one wouldn't notice anything if all the pieces of the puzzle are in place - but it seems we use "involuntary" a single puzzle piece and hence wonder about the big picture.

> How do you use forms in Pimcore/Symfony?

I don't even know how to answer that, because as far as I'm aware we use pretty much plain Symfony forms for the most part.
```
        $form = $this->createForm(LostPasswordFormType::class, [],[
            'action' => $this->generateUrl('user_lostpassword'),
            'userProvider' => $userProvider,
        ]);
        $form->handleRequest($request);
```
And the form type itself uses the Symfony core Form types like submit and text and implements some constraints - nothing really fancy in terms of form handling.
Only special thing is that it has _no_ BlockPrefix aka "form name", and this is exactly the point where the adjusted form handler trips up.
We have a POST form, but the adjusted FormHandler doesn't check that (on purpose, but I don't know why) and hence simply submits empty data in case of a form without a blockPrefix / name.

For now I've overwritten the service definition to get rid of the decorator.
This patch is just in place in case it turns out there's actually some dependency that needs this special behavior later on.

dpfaffenbauer:
and there is the difference. I have to look closer into it, but our ResourceController, which handles all backed types for adding/editing/deleting Resources, is quite generic. It uses forms for basically just editing. But we plan on using it also for adding, to have better input validations. All of these requests gonna be POST requests. So we might be able to remove the handler. Needs tests though

In the meantime, this is how I handle symfony forms:

```
$form = $this->formFactory->createNamed('contact', ContactFormType::class);

if ($info->getRequest()->isMethod('post')) {
     $form->handleRequest($info->getRequest());
...
}
```

That is also the difference between CoreShop's Request Form handler, and the symfony handler. Our's doesn't check for Request method, whereas symfony's does. Example that would work with CoreShop's Request Handler.

```
$form = $this->formFactory->createNamed('contact', ContactFormType::class, null, ['method' => 'post']);
$form->handleRequest($info->getRequest());
```

So, for now to solve your issue:

 - Check the request before submitting the form
 - Can you also create a ticket if it is possible to remove the `HttpFoundationRequestHandler`?

PR #1583, opened 2021-01-31
Composer compatibility with Coreshop 2.1 series

Cruiser13:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Solves errors like Class 'CoreShop\Component\Address\Model\StateInterface' does not exist after a composer update.

I know that the 2.1 series is old but this could be worth a release in my opinion.

### DIFF
--- a/src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php
+++ b/src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/DoctrineTargetEntitiesResolverPass.php
@@ -6,12 +6,14 @@
  * For the full copyright and license information, please view the LICENSE.md and gpl-3.0.txt
  * files that are distributed with this source code.
  *
- * @copyright  Copyright (c) 2015-2020 Dominik Pfaffenbauer (https://www.pfaffenbauer.at)
+ * @copyright  Copyright (c) 2015-2021 Dominik Pfaffenbauer (https://www.pfaffenbauer.at)
  * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
  */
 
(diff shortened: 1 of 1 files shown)


Cruiser13:
We'd actually need to also remove Versions::VERSIONS constant as it's no longer available. Not sure how the getComposerPackageName functions in CoreShopFrontendBundle.php and CoreShopIndexBundle and so on are used but this needs change too.

dpfaffenbauer:
CoreShop 2.1 does not receive updates anymore, if you need it, use Composer v1

dpfaffenbauer:
also the issue with StateInterface goes not get merged into 2.1. downgrade doctrine to make it work...

Cruiser13:
@dpfaffenbauer there are hosts which will auto update composer. Not having the 2.1 coreshop work in these environments is a huge issue. If downgrading doctrine will work, could we add that to the docs somewhere?

solverat:
@Cruiser13 really? You should consider switching such providers then. Imagine older installations with unstable/incomplete psr-4 standard. They simply can't update, and they simply will break after a composer install on v2.

Cruiser13:
@solverat that's sadly nothing I can personally change. Out of my hands.

Cruiser13:
> 
> 
> CoreShop 2.1 does not receive updates anymore, if you need it, use Composer v1

I'd not create new features for 2.1 but breaking changes should be addressed I think. Updating 2.1 to 2.2 is error prone so there are a lot of 2.1 versions out there. They'll stop working with the next composer update, even if they do not update coreshop itself.

dpfaffenbauer:
@Cruiser13 no, simply not. I don't have the time or the resources to support older versions.  Sorry, but that is out of my hands.

PR #1588, opened 2021-02-15
[Core] remove country-aware currency context as it messes with the Storage Based Currency Context

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no


solverat:
to reproduce:
- allow two currencies: one store "test" with base currency `EUR` and an additional currency `CHF`
- go to `GeoLiteBasedRequestResolver` 
- add on top of method `findCountry`:  `return $this->countryRepository->findByCode("CH");`
- go to frontend for store "test": `CHF` is selected

=> incorrect - should be `EUR` until changes currency by triggering the `switch-currency` route.

solverat (dismissed):

Issue #1602, opened 2021-03-23
[IndexBundle] Range Filter does not work if min or max value is 0


ychanan:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

If min value (and/or max value) in the range is zero (e.g. 0-500), the RangeFilterConditionProcessor skips applying the condition to the list.

As seen in the `RangeFilterConditionProcessor::addCondition()` method:
```
if (!empty($valueMin) && !empty($valueMax)) {
    $fieldName = $field;

    if ($isPrecondition) {
        $fieldName = 'PRECONDITION_' . $fieldName;
    }

    $list->addCondition(new RangeCondition($field, $valueMin, $valueMax), $fieldName);
}
```

dkarlovi:
Right, I don't see why 0 should be anything special, min can easily be negative, the range should check for `!== null` probably.

PR #1606, opened 2021-04-05
[IndexBundle] Fix range filter to work if min/max value = 0

ychanan:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1602

### DIFF
--- a/src/CoreShop/Component/Index/Filter/RangeFilterConditionProcessor.php
+++ b/src/CoreShop/Component/Index/Filter/RangeFilterConditionProcessor.php
@@ -32,13 +32,23 @@ class RangeFilterConditionProcessor implements FilterConditionProcessorInterface
         $minValue = count($rawValues) > 0 ? $rawValues[0]['value'] : 0;
         $maxValue = count($rawValues) > 0 ? $rawValues[count($rawValues) - 1]['value'] : 0;
 
+        $currentValueMin = $minValue;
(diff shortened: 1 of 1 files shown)


src/CoreShop/Component/Index/Filter/RangeFilterConditionProcessor.php
```diff
@@ -37,8 +37,8 @@ public function prepareValuesForRendering(FilterConditionInterface $condition, F
             'label' => $condition->getLabel(),
             'minValue' => $minValue,
             'maxValue' => $maxValue,
-            'currentValueMin' => $currentFilter[$field . '-min'] ? $currentFilter[$field . '-min'] : $minValue,
-            'currentValueMax' => $currentFilter[$field . '-max'] ? $currentFilter[$field . '-max'] : $maxValue,
+            'currentValueMin' => ('' !== (string)$currentFilter[$field . '-min']) ? $currentFilter[$field . '-min'] : $minValue,
```
dpfaffenbauer, line 40: I think we should check if the index `$currentFilter[$field . '-min']` actually is set here? PHP with strict-types would trigger a warning here

ychanan: Resolved in a new commit.

dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1645, opened 2021-05-26
Fix Firewall for URL path /

BlackbitNeueMedien:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When the core shop is accessible directly under URL path `/` (so like when you call the home page https://example.org/) the logged in customer cannot be found with the `TokenBasedRequestResolver` because path `/` does not match the regex `^/(?!admin)[^/]++`.

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -1,5 +1,5 @@
 parameters:
-    coreshop.security.frontend_regex: "^/(?!admin)[^/]++"
+    coreshop.security.frontend_regex: "^/(?!admin)[^/]*"
 
 security:
     providers:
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
I guess this seems alright. I just approved the tests to run. should run fine though.

BTW: You know this change then only affects CoreShop 3.0 where CoreShop itself doesn't control the firewall anymore?

dpfaffenbauer (approved):

dpfaffenbauer:
tests fail for other reasons

dpfaffenbauer:
thanks

Issue #1618, opened 2021-04-27
[QPR] Cannot change Conditions


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Pimcore: >= 6.8, reproducible on demo2

- Create QPR
- Add condition (store, timespan)
- Add range
- Save
- Reopen
- Change one of given conditions)
- Save
- Reopen

-> Check Condition: no changes persisted

solverat:
Also: Condition does not get removed from `coreshop_rule_condition` table even if the relations has been removed from the `coreshop_product_quantity_price_rule_conditions` table.

Issue #1557, opened 2021-01-18
Notification Rule - not able to send more then one email


albertmueller:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes/no

Tryed to set more then one email template in notification rule, what happen if you press save:
![image](https://user-images.githubusercontent.com/2986829/104914109-efe62e80-598e-11eb-9347-d6ca5555ad64.png)

After a look in the db table: coreshop_rule_action, it shows the same id from the latest template in both entry configuration.

ramundomario:
Tested it and the email is sent 2 times with the same mail document (unless you manually change the id from the database)

My use case: Send order confirmation email to both the customer and the store team

![Screenshot from 2021-02-12 14-27-16](https://user-images.githubusercontent.com/16133208/107773783-71d33880-6d3e-11eb-9e13-4ae95e085f75.png)
The last mail document (id: 12) is used as mail in both actions as visible from the database:
![Screenshot from 2021-02-12 14-23-38](https://user-images.githubusercontent.com/16133208/107773787-739cfc00-6d3e-11eb-8d9c-6a4296016662.png)

Reproduced on [latest demo](http://demo2.coreshop.org/)

dpfaffenbauer:
guess that is fixed with #1657

PR #1657, opened 2021-06-14
[Rule] refactor Rule conditions and actions peristance

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Fixed tickets | #1618 

THIS IS A BC BREAK! But I don't see a way around it. This is due to how Symfony Form works with Collections/Arrays. We have to refactor that in order to make it properly work.

 - Add sorting field
 - Fixes in Entity Merger with ManyToMany Collections
 - Add reloading of conditions/actions on save
 - Force reload Resource on save
 - Fix orphan removal for conditions and actions

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Migrations/Version20210614115749.php
+++ b/src/CoreShop/Bundle/CoreBundle/Migrations/Version20210614115749.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace CoreShop\Bundle\CoreBundle\Migrations;
+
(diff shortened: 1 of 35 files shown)


ramundomario:
Could also solve #1557?

dpfaffenbauer:
@ramundomario probably

dpfaffenbauer:
@ramundomario I just tested it with this fix, it works for sending different emails. But I never tested it before the fix.

PR #1658, opened 2021-06-17
Controller error with Pimcore X

bramalho:
This PR solves the issue https://github.com/coreshop/CoreShop/issues/1659

`Resources/config/services.yml` and `Resources/config/services/dynamic_dropdown.yml` auto-configures and registers the admin controllers.

This solves the error:
```
"CoreShop\Bundle\PimcoreBundle\Controller\Admin\DynamicDropdownController" has no container set, did you forget to define it as a service subscriber?
```

`Controller/Admin/DynamicDropdownController.php`
If the class is an instance of `Pimcore\Model\DataObject\ClassDefinition`, get the layout definition.
This solves the error:
```
Warning: Undefined array key "getName"

in vendor/coreshop/pimcore-bundle/Controller/Admin/DynamicDropdownController.php (line 183)
        $tree = include $definitionFile;
        $definition = $this->parseTree($tree, []);

        return $definition[$method];
```
when trying to get some value from the assigned class.


src/CoreShop/Bundle/PimcoreBundle/Resources/config/services.yml
```diff
@@ -11,8 +11,15 @@ parameters:
 
 services:
     _defaults:
+        autowire: true
+        autoconfigure: true
         public: true
 
+    CoreShop\Bundle\PimcoreBundle\Controller\Admin\:
```
dpfaffenbauer, line 18: within the coreshop-core we don't use auto-configuration/auto-wiring

dpfaffenbauer (changes requested):

src/CoreShop/Bundle/PimcoreBundle/Resources/config/services.yml
bramalho: auto-wire needs to be done for the Admin controller, added only for this controller in `src/CoreShop/Bundle/PimcoreBundle/Resources/config/services/dynamic_dropdown.yml`

src/CoreShop/Bundle/PimcoreBundle/Resources/config/services/dynamic_dropdown.yml
```diff
@@ -4,5 +4,6 @@ services:
         public: true
 
     CoreShop\Bundle\PimcoreBundle\Controller\Admin\DynamicDropdownController:
+        autowire: true
```
dpfaffenbauer, line 7: please remove

PR #1661, opened 2021-06-24
storeId should be surrounded by commas in searchController

seybsen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

storeId should be surrounded by commas to avoid matching for example storeId `10` when searching in store `1`


dpfaffenbauer:
we don't accept PR's into 2.2-dev-3.0. that is a temporary branch for some projects to run the dev version of coreshop 3 on pimcore 6

PR #1703, opened 2021-07-27
Prevent PHP exception by non nullable return type

paulverdu:
When creating an object the $id was overwritten in the getResourceId method and set to null. This created a php exception and breaks the code.
This code change will prevent that by checking if the id of the resource object is not null.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -


dpfaffenbauer:
thanks

paulverdu:
@dpfaffenbauer When do you expect to release this fix, so it comes available in data-definitions? This kinda breaks data definitions now since a new ImportDefinition can't be created.

dpfaffenbauer:
@paulverdu I just made the split of the repos. If you run composer-update you should get the version with this fix

PR #1706, opened 2021-07-28
exchange fzaninotto/faker, change with fakerphp/faker -- 2.2 dev 3.0

RobbiXYZ:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Fixed tickets | 

Request: Please take out `fzaninotto/faker`, possibly change with `fakerphp/faker`, or other.

Because `fzaninotto/faker` did not like php 8,  
```Workflow: Install 
        1. Install pimcore and 
        2. Install coreshop
```
is not possible for me, because of `fzaninotto/faker`.  Exchange not possible via composer, since it is not a direct dependency [Source: Julien B.](https://stackoverflow.com/users/4346453/julien-b).

My unsuccessful try's to exchange via composer I have described on [stackoverflow](https://stackoverflow.com/questions/68559676/uninstall-fzaninotto-faker-and-change-with-fakerphp-faker-via-composer).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -193,7 +193,7 @@
         "symfony-tests-dir": "tests",
         "symfony-assets-install": "relative",
         "branch-alias": {
-            "dev-master": "2.4-dev"
+            "dev-2.2-dev-3.0": "2.4-dev"
         }
     }
 }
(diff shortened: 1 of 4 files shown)


dpfaffenbauer:
nothing will get merged into 2.2-dev-3.0

PR #1715, opened 2021-08-17
[Tracking] add docu for setting storeValues

lukadschaak:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Hey Dominik, can you please take a look, if this way of updating storeValues is intended, or is it crap?
I wrote it in style of documentation, so when the code may be alright it can used later by other people.

I found no docu about the store values at all.

### DIFF
--- a/docs/03_Development/13_Ecommerce_Tracking/01_Product_Prices
+++ b/docs/03_Development/13_Ecommerce_Tracking/01_Product_Prices
@@ -1,0 +1,59 @@
+# Product Prices in Tracking
+
+When a product impression is tracked, the price for the product will be read from the storeValues, a data field in the class definition of the CoreShopProduct.
+
+To have prices in the tracking events, the storeValues must be maintained. This can be done for example with a subscriber as shown below.
+
+The real price must be determined by suitable service.
+
+```php
+<?php declare(strict_types=1);
+
+namespace AppBundle\CoreShop\EventListener;
+
+use Doctrine\Persistence\ObjectRepository;
+use Pimcore\Event\DataObjectEvents;
+use Pimcore\Event\Model\DataObjectEvent;
+use Pimcore\Model\DataObject\CoreShopProduct;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class StoreValuesProductUpdateSubscriber implements EventSubscriberInterface
+{
+    private const REPLACE_THIS_EXAMPLE_PRICE_WITH_SERVICE_CALL = 1234;
+
+    private ObjectRepository $storeRepository;
+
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
settting the store values is right, but I don't see the point about it with tracking? I mean, if you want to sell a product, it should have a price...

lukadschaak:
We configure our prices via the tab "Price Rules". Per product we create a quantity price rule with a Stückmengenpreisregel from 1 piece. But this price is not stored in the store values.
Are our prices maybe misconfigured?

-- german version
Wir setzen unsere Preis über den Reiter Price Rules. Wir haben pro Product eine Quantity Price Rule mit einer Stückmengen Preisregel ab 1 Einheit. Aber dieser Preis taucht nicht in den storeValues auf.
Ich war davon ausgegangen, dass dieser Preis auch in die storeValues geschrieben wird, aber dort steht nichts. Vielleicht haben wir unsere Preise falsch konfiguriert?!

grafik
grafik
grafik

dpfaffenbauer:
this is one possible way of configuring it... coreshop has many use case for things like that.

what is the effect you experience with not having a store-price? you don't get a price at all in tracking?

lukadschaak:
Exactly. The ProductPriceCalculator returns an error for not having store values for the requested product (there are no store values at all). And theirfor the price is always 0 as fallback.

Currently i try to find out how the price is determined for cartItems. Because there was never a problem with having no prices.

dpfaffenbauer:
I see. The ProductPriceCalculator is one of the price-calculators. But: Quantity Price Rules are different, they don't get calculated for a product, but only for the cart. Usually you would have a "default" price in store values, and all quantity-price-rules starting with a special discount. Like default price is 10€, but starting from 10, it is 9€.

lukadschaak:
Ahh, that would make sense.
After a bit of research: Our Team get the prices for a product by creating temporary cartItems, push them to the cart and using the '@coreshop.cart_processor' to calculate prices. And then it happens what you said before. For carts, the quantity-price-rules are used in the caluculation.

So what would be best practice to create the default values for a store? Simply setting the values via PHP Code (`$product->setStoreValuesOfType('price', 1234, $store);`)?

But then the price is not visible for pimcore backend user. Is there a way of setting the default price per store via the pimcore backend? Or at least the default price for all stores :)

dpfaffenbauer:
for tracking you should extend the Trackings ProductExtractor to do what you already do with the price display.

lukadschaak:
Ok, thanks :+1:

PR #1722, opened 2021-09-15
fix: Error on saving money field

marlonbasten:
When saving a DataObject containing a money field, it errors out on saving, because the number typed in into the field is parsed as a string to the PHP function. That is easily mitigated by parsing the string into a float, because the function already checks if the string is numeric.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

[This](https://prnt.sc/1sa6csc) happens normally when trying to save a DataObject containing a CoreShop money field.


dpfaffenbauer (approved):

dpfaffenbauer:
thasnk

Issue #1765, opened 2021-11-02
Duplicate Entry Exception when Copy/Paste Product


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

- Create Product
- Set default unit definition
- save and publish
- copy object
- past object:

![image](https://user-images.githubusercontent.com/700119/139847946-7ac5a9fa-941a-4ccb-a90d-7ec88eb10b58.png)

![image](https://user-images.githubusercontent.com/700119/139847781-70c3b4fb-6813-4d2c-a9d4-975fefce4fa8.png)

dpfaffenbauer:
Version?

solverat:
2.2

solverat:
also on x. reproducable on demox.

ramundomario:
On demox, when copy/paste, there is an additional error regarding uniqueness of slug:
![Screenshot from 2021-11-02 14-11-06](https://user-images.githubusercontent.com/16133208/139853409-94d63906-61d8-4598-95f6-4b62efed2f7f.png)

dpfaffenbauer:
@ramundomario Slug hast been solved already. The SluggableListener now checks for uniqueness and adds a -1 if one already exists.

PR #1770, opened 2021-11-04
[CoreExtensions] refactor how Doctrine Entities are cloned

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #1765

Don't use `__clone` anymore, as it is way too unpredictable in some matters. Use a `CustomCopy` from Pimcore and manually clone and reset data. This also doesn't completely clone all the data, since we have some dependencies from store-values to Product Unit Definitions. But this is something we cannot solve this way and therefore is simply just not supported. This also fixes copying of Specific Product Price Rules and Quantity Price Rules.

<!-

### DIFF
--- a/src/CoreShop/Component/Pimcore/BCLayer/CustomDataCopyInterface.php
+++ b/src/CoreShop/Component/Pimcore/BCLayer/CustomDataCopyInterface.php
@@ -12,7 +12,7 @@
 
 namespace CoreShop\Component\Pimcore\BCLayer;
 
-if (interface_exists(\Pimcore\Model\DataObject\ClassDefinition\Data\CustomVersionMarshalInterface::class)) {
+if (interface_exists(\Pimcore\Model\DataObject\ClassDefinition\Data\CustomDataCopyInterface::class)) {
     interface CustomDataCopyInterface extends \Pimcore\Model\DataObject\ClassDefinition\Data\CustomDataCopyInterface
     {
     }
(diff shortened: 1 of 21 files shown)


src/CoreShop/Component/Pimcore/BCLayer/CustomDataCopyInterface.php
```diff
@@ -0,0 +1,23 @@
...
+ * files that are distributed with this source code.
+ *
+ * @copyright  Copyright (c) CoreShop GmbH (https://www.coreshop.org)
+ * @license    https://www.coreshop.org/license     GNU General Public License version 3 (GPLv3)
+ */
+
+namespace CoreShop\Component\Pimcore\BCLayer;
+
+if (interface_exists(\Pimcore\Model\DataObject\ClassDefinition\Data\CustomVersionMarshalInterface::class)) {
+    interface CustomDataCopyInterface extends \Pimcore\Model\DataObject\ClassDefinition\Data\CustomDataCopyInterface
```
jdreesen, line 16: Is it intended that the `interface_exists()` call checks a different class than the one that's extended?

dpfaffenbauer: thanks

dpfaffenbauer:
ping @solverat

PR #1773, opened 2021-11-09
[Store select / multiselect] Support getOptions() via option provider

BlackbitNeueMedien:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Currently it is not possible to fetch the available options for store select / multiselect field types as you would normally do for Pimcore default select / multiselect fields via 
```php
$object->getClass()->getFieldDefinition('store')->getOptions(); // returns always null
DataObject\Service::getOptionsForSelectField($object, "store"); // also returns always null
```
This is currently always `null` as the logic for fetching the available options is only implemented in JS, see https://github.com/coreshop/CoreShop/blob/99bbb0c927d2cc57ca81bd3b03dbd5805a6172c8/src/CoreShop/Bundle/StoreBundle/Resources/public/pimcore/js/coreExtension/tags/coreShopStoreMultiselect.js#L17 and https://github.com/coreshop/CoreShop/blob/99bbb0c927d2cc57ca81bd3b03dbd5805a6172c8/src/CoreShop/Bundle/ResourceBundle/Resources/public/pimcore/js/coreExtension/tags/multiselect.js#L23

It works currently only because Pimcore does not check if the set option is actually available in `Select::checkValidity()` / `Multiselect::checkValidity()` in https://github.com/pimcore/pimcore/blob/d0b5dc0d78392ee2d4012d228214901fd1080076/models/DataObject/ClassDefinition/Data/Select.php#L373
But this may change and nevertheless it is a good idea to support the same logic like Pimcore's default field types do.

### DIFF
--- a/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
+++ b/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
@@ -51,6 +51,6 @@ class Store extends Select
 
     public function getOptionsProviderClass()
     {
-        return StoreOptionProvider::class;
+        return '@'.StoreOptionProvider::class;
     }
 }
--- a/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
+++ b/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
@@ -48,4 +48,9 @@ class Store extends Select
     {
         return true;
     }
+
+    public function getOptionsProviderClass()
+    {
+        return StoreOptionProvider::class;
+    }
 }
--- a/src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreMultiselect.php
+++ b/src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreMultiselect.php
@@ -27,4 +27,9 @@ class StoreMultiselect extends Multiselect
      * @var string
      */
     public $fieldtype = 'coreShopStoreMultiselect';
+
+    public function getOptionsProviderClass()
+    {
+        return '@'.StoreOptionProvider::class;
+    }
 }
--- a/src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreOptionProvider.php
+++ b/src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreOptionProvider.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * CoreShop.
+ *
+ * This source file is subject to the GNU General Public License version 3 (GPLv3)
+ * For the full copyright and license information, please view the LICENSE.md and gpl-3.0.txt
+ * files that are distributed with this source code.
+ *
+ * @copyright  Copyright (c) CoreShop GmbH (https://www.coreshop.org)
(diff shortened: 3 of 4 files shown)


src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreOptionProvider.php
```diff
@@ -0,0 +1,71 @@
...
+
+class StoreOptionProvider implements SelectOptionsProviderInterface
+{
+    /** @var StoreRepository */
+    private $repository;
+
+    /**
+     * @param StoreRepository $repository
+     */
+    public function __construct(StoreRepository $repository)
```
dpfaffenbauer, line 27: use PHP8 Constructor property promotion

src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreMultiselect.php
```diff
@@ -27,4 +27,9 @@ class StoreMultiselect extends Multiselect
      * @var string
      */
     public $fieldtype = 'coreShopStoreMultiselect';
+
+    public function getOptionsProviderClass()
+    {
+        return '@'.StoreOptionProvider::class;
```
dpfaffenbauer, line 33: '@'?

src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
```diff
@@ -48,4 +48,9 @@ protected function getNullable(): bool
     {
         return true;
     }
+
+    public function getOptionsProviderClass()
+    {
+        return StoreOptionProvider::class;
```
dpfaffenbauer, line 54: no '@'?

dpfaffenbauer (changes requested):

src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreOptionProvider.php
BlackbitNeueMedien: Fixed. I do not like this syntax but doesn't matter...

src/CoreShop/Bundle/StoreBundle/CoreExtension/StoreMultiselect.php
BlackbitNeueMedien: See https://github.com/pimcore/pimcore/blob/4550a053fac47a1d2c0406bf15c9799ef078bbb7/models/DataObject/ClassDefinition/Helper/ClassResolver.php#L35-L40
Because we need dependency injection here, we want to use the symfony service. And this is done with the prepending `@`.

src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
BlackbitNeueMedien: Fixed.

dpfaffenbauer (approved):

dpfaffenbauer:
PSALM is failing for other reasons. Interesting that it does that all the time. it passes and a bit later finds new issues :/. 

But has nothing todo with your pr. so thanks for your contribution.

Issue #1699, opened 2021-07-23
CustomerAndStoreBasedCartContext does not respect RememberMe Dispatch


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? |no
| BC Break report? |no
| RFC?             | no

Because of this:
https://github.com/coreshop/CoreShop/blob/002f9cf782c9d647174a38bac638b37a9eae1db8/src/CoreShop/Bundle/OrderBundle/Context/CustomerAndStoreBasedCartContext.php#L49-L50

...which has been introduced in https://github.com/coreshop/CoreShop/pull/340.

⚠️ **This route check is still valid and cannot be changed otherwise issues like described in #340 will reoccur.** 

***

Also: There should **not**  be an exact result check  in `findLatestByStoreAndCustomer` => `count($objects) >= 1` should be enough. Otherwise, the cart blamer will never find a cart again if two or more are available:

https://github.com/coreshop/CoreShop/blob/002f9cf782c9d647174a38bac638b37a9eae1db8/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php#L65-L66

***

## Reproduction
- Login and hit "remember me"
- Add some products to cart
- Close Browser (Completely)
- Reopen Browser, open Store
- Cart is empty

## Issue
A lot of new persisted carts and weird user experiences (Empty cart after re-visiting store).

## Solution
Not easy. I had to do a quick fix by setting a request attribute in `CartBlamerListener::onInteractiveLogin()`:

```php
$force = $interactiveLoginEvent->getAuthenticationToken() instanceof RememberMeToken;
$interactiveLoginEvent->getRequest()->attributes->set('coreshop.force_cart_blame', $force);
```

Then I added a custom CartContext which will dispatch the `$this->cartRepository->findLatestByStoreAndCustomer($store, $customer);` if this attribute is available in the request.
Since we cannot pass any parameters via `cartContext->getCart()` this seems to be the only solution by now.

dpfaffenbauer:
What was the use case https://github.com/coreshop/CoreShop/pull/340 again? How does a user come back to the Site and is still in the Order process without actually coming back to the revise page?

solverat:
- add items as guest
- log in ⇒ guest cart will be attached to the user (1)
- log off
- add items as guest
- log in ⇒ guest cart will be attached to the user (2)
- checkout and complete purchase
- cart (1) gets attached

That's why it should stay that way.

[...]

PR #1779, opened 2021-11-11
[Cart] fix existing cart initalization on customer login

albertmueller:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

specially on findLatestByStoreAndCustomer it can happen that more then one cart exists for the customer, when the user has logged in, he has been shown an empty cart. although several carts exist, the last cart should be loaded.

### DIFF
--- a/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
+++ b/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
@@ -41,6 +41,7 @@ class CartRepository extends PimcoreRepository implements CartRepositoryInterfac
     {
         $list = $this->getList();
         $list->setCondition('customer__id = ? AND name = ? AND order__id is null', [$customer->getId(), $name]);
+        $list->setLimit(1);
         $list->load();
 
         $objects = $list->getObjects();
@@ -58,6 +59,7 @@ class CartRepository extends PimcoreRepository implements CartRepositoryInterfac
         $list->setCondition('customer__id = ? AND store = ? AND order__id is null ', [$customer->getId(), $store->getId()]);
         $list->setOrderKey('o_creationDate');
         $list->setOrder('DESC');
+        $list->setLimit(1);
         $list->load();
 
         $objects = $list->getObjects();
@@ -73,6 +75,7 @@ class CartRepository extends PimcoreRepository implements CartRepositoryInterfac
     {
         $list = $this->getList();
         $list->setCondition('o_id = ? AND order__id is null ', [$id]);
+        $list->setLimit(1);
         $list->load();
 
         $objects = $list->getObjects();
--- a/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
+++ b/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
@@ -37,22 +37,6 @@ class CartRepository extends PimcoreRepository implements CartRepositoryInterfac
         return $carts;
     }
 
-    public function findNamedForCustomer(CustomerInterface $customer, $name): ?OrderInterface
-    {
-        $list = $this->getList();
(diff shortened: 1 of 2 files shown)


solverat:
also see #1699

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
```diff
@@ -73,6 +75,7 @@ public function findCartById($id): ?OrderInterface
     {
         $list = $this->getList();
         $list->setCondition('o_id = ? AND order__id is null ', [$id]);
+        $list->setLimit(1);
```
dpfaffenbauer, line 78: that should always ever return one or null since we directly filter on o_id.

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
```diff
@@ -41,6 +41,7 @@ public function findNamedForCustomer(CustomerInterface $customer, $name): ?Order
     {
         $list = $this->getList();
         $list->setCondition('customer__id = ? AND name = ? AND order__id is null', [$customer->getId(), $name]);
+        $list->setLimit(1);
```
dpfaffenbauer, line 44: is not right here, named can return multiple carts

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
```diff
@@ -58,6 +59,7 @@ public function findLatestByStoreAndCustomer(StoreInterface $store, CustomerInte
         $list->setCondition('customer__id = ? AND store = ? AND order__id is null ', [$customer->getId(), $store->getId()]);
         $list->setOrderKey('o_creationDate');
         $list->setOrder('DESC');
+        $list->setLimit(1);
```
dpfaffenbauer, line 62: that is right

dpfaffenbauer (changes requested):

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
albertmueller: yes, but as far as i know, the query is faster with a limit, even if you won't notice it in reality :)

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
albertmueller: hm, but then is the following code below wrong:
`if (count($objects) === 1 && $objects[0] instanceof OrderInterface) {
            return $objects[0];
        }`
because only one cart will returned?

albertmueller: is this function somewhere in use @dpfaffenbauer? i can't find a usage in coreshop, if not then how about to return the found collection of carts?

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
dpfaffenbauer: I am not sure, it feels wrong here, so please remove it

src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/CartRepository.php
dpfaffenbauer: true, it always ever return if it is one. We don't have any usage of it CoreShop. I don't think it even works... So we should remove it with 3.0

dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1797, opened 2021-11-23
[Tracking] Composite increase performance

### LINKED ISSUE #1728: [Bug][perf] CompositeTracker collects product data even if no actual tracker is enabled
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

I did something like this:

```php
    public function isEnabled()
    {
        if ($this->enabled !== null) {
            return $this->enabled;
        }
        
        foreach ($this->trackerRegistry as $tracker) {
            if ($tracker->isEnabled()) {
                $this->enabled = true;
                
                return true;
            }
        }
        
        return $this->enabled = false;
    }
    ```
    
Then each call to `track*()` in the composite tracker checks 

```php
        if (!$this->isEnabled()) {
            return;
        }
```

skipping an expensive `extractTrackingData` which is not used.

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | fixes #1728

@dkarlovi like this?

### DIFF
--- a/src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
+++ b/src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
@@ -33,9 +33,7 @@ class CompositeTracker implements TrackerInterface
 
         foreach ($this->trackerRegistry as $tracker) {
             if ($tracker->isEnabled()) {
-                $this->enabled = true;
-
-                return true;
+                return $this->enabled = true;
             }
         }
 
--- a/src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
+++ b/src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
@@ -19,13 +19,27 @@ use CoreShop\Component\Tracking\Extractor\TrackingExtractorInterface;
 
 class CompositeTracker implements TrackerInterface
 {
+    private ?bool $enabled = null;
+
     public function __construct(private TrackingExtractorInterface $extractor, private ServiceRegistryInterface $trackerRegistry)
     {
     }
 
     public function isEnabled(): bool
     {
-        return true;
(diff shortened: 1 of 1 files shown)


dkarlovi (approved):
LGTM.

src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
```diff
@@ -19,13 +19,27 @@
...
-        return true;
+        if ($this->enabled !== null) {
+            return $this->enabled;
+        }
+
+        foreach ($this->trackerRegistry as $tracker) {
+            if ($tracker->isEnabled()) {
+                $this->enabled = true;
+
+                return true;
```
jdreesen, line 38:
```suggestion
                return $this->enabled = true;
```

src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
```diff
@@ -31,7 +31,7 @@ public function isEnabled(): bool
             return $this->enabled;
         }
 
-        foreach ($this->trackerRegistry as $tracker) {
+        foreach ($this->trackerRegistry->all() as $tracker) {
```
dkarlovi, line 34: SA, the most boring colleague you want to have. :)

dkarlovi (approved):

src/CoreShop/Component/Tracking/Tracker/CompositeTracker.php
dpfaffenbauer: ;)

PR #1807, opened 2021-12-02
fix new form namespace

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

This is an regression of https://github.com/coreshop/CoreShop/pull/1597. The checkout forms does now comes with a namespace "coreshop". So `checkout_finisher` is not in range anymore. This PR fixes that.

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Resources/views/Checkout/steps/summary.html.twig
+++ b/src/CoreShop/Bundle/FrontendBundle/Resources/views/Checkout/steps/summary.html.twig
@@ -20,10 +20,10 @@
 
         <div class="row">
             <div class="col-12 col-sm-1 offset-sm-10">
-                <button type="submit" name="checkout_finisher" value="{{ coreshop_path(cart, 'coreshop_cart_create_qoute') }}" class="btn btn-success w-100">{{ 'coreshop.ui.quote'|trans }}</button>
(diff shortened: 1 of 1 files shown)


solverat:
There is more. With that, the response fragment is not valid anymore:

https://github.com/coreshop/CoreShop/blob/27c96881a44596b5b1d12e4d36b2c133c99b7494/src/CoreShop/Bundle/CoreBundle/Checkout/Step/SummaryCheckoutStep.php#L45-L47

Since this is now an array (`$request->get('coreshop')['checkout_finisher'];`). Not sure how to handle this. Removing the `coreshop` namespace from checkout steps could be a solution but is also inconsistent.

ramundomario:
I reproduced too. Basically when terms are not accepted, form appear as not submitted and validation error is not shown.
Instead, if terms are accepted, the form appear submitted and everything works fine.
Let's wait thoughts from @dpfaffenbauer about get url from response.

ramundomario:
> There is more. With that, the response fragment is not valid anymore:
> 
> https://github.com/coreshop/CoreShop/blob/27c96881a44596b5b1d12e4d36b2c133c99b7494/src/CoreShop/Bundle/CoreBundle/Checkout/Step/SummaryCheckoutStep.php#L45-L47
> 
> Since this is now an array (`$request->get('coreshop')['checkout_finisher'];`). Not sure how to handle this. Removing the `coreshop` namespace from checkout steps could be a solution but is also inconsistent.

Reminder, currently it is not possible to get to the payment step @dpfaffenbauer

dpfaffenbauer:
Can you please open a separate story for that?

PR #1808, opened 2021-12-03
Allow null as Parameter

lukas-schnieper:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Allow null as parameter, because that is what happens if you access the shop without an existing cart or if you try to add a new product unit definition.

### DIFF
--- a/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/OrderRepository.php
+++ b/src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/OrderRepository.php
@@ -40,7 +40,7 @@ class OrderRepository extends PimcoreRepository implements OrderRepositoryInterf
         return $carts;
     }
 
-    public function findByCartId(int $id): ?OrderInterface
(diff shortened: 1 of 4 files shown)


src/CoreShop/Bundle/OrderBundle/Pimcore/Repository/OrderRepository.php
```diff
@@ -40,7 +40,7 @@ public function findCartByCustomer(CustomerInterface $customer): array
         return $carts;
     }
 
-    public function findByCartId(int $id): ?OrderInterface
+    public function findByCartId(?int $id): ?OrderInterface
```
jdreesen, line 43: Maybe you should return early in this case, as `o_id` afaik cannot be `null` and we don't need to query the database in this case.

dpfaffenbauer:
something is wrong on other places, this is good as it is.

PR #1814, opened 2021-12-09
[CoreBundle] Fix fetch of `checkout_finisher` url from request

ramundomario:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   

Followup https://github.com/coreshop/CoreShop/pull/1807#issuecomment-989913502

It should fix the Behat UI failed test of #1812

Instead of taking the value from request we could get the value from the extra data of the form, let me know WDYT @dpfaffenbauer


dpfaffenbauer:
@ramundomario I thought about adding the submit buttons to the form and take it from there. You have another idea?

dpfaffenbauer (approved):

dpfaffenbauer:
thanks btw :)

dpfaffenbauer:
@ramundomario definitely. We cannot change that for 2.x, but we can do it for 3.x. Would you create a PR?

ramundomario:
@dpfaffenbauer I'll try, maybe in the next week.

Issue #1746, opened 2021-10-18
Checkout :: addresss :: Add new billing/shipping address not working


ivicafactory:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

On checkout, as a logged-in user, I want to add a new billing address. The expected result is that I can enter that address and what happens is that I get redirected to the homepage. The same issue is with the shipping address.

Video: https://watch.screencastify.com/v/Q3qnkXSgyd1G3UYYLTKS

dpfaffenbauer:
Interesting, it doesn't appear that you are logged in. Is it possible you used guest checkout?

dpfaffenbauer:
does it still happen for you @ivicafactory ?

valentin-:
@dpfaffenbauer when adding a new address from the checkout or in the customer address settings as logged in user, I get an error as well.

In my case `$address->getParent()` is null and therefore throws an error on `getId()`

https://github.com/coreshop/CoreShop/blob/94c133f65a704a8486e9bb1cd9687f852028e937/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php#L129

m0nken:
I also ran into this. As this occurred when I tried adding a new address, I would think the problem lies within `$this->isNewEntity($address)` which should return true for a newly added address and therefore the second part `$address->getParent()->getId() !== $newParent->getId()` should not be called at all in this case.

m0nken:
Nevermind. I just saw that this has already been addressed [here](https://github.com/coreshop/CoreShop/commit/de54b6069c416edda5784fa384a6a020d9eed936).

PR #1825, opened 2021-12-23
Add address parent check

valentin-:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | https://github.com/coreshop/CoreShop/issues/1746

Add a check if the parent exists. Tested in /de/shop/customer/addresses and also in the checkout process.

### DIFF
--- a/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php
+++ b/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php
@@ -126,7 +126,7 @@ final class CustomerTransformHelper implements CustomerTransformHelperInterface
         }
 
         // set new or changed parent
-        if ($this->isNewEntity($address) === true || ($address->getParent() && $address->getParent()->getId() !== $newParent->getId())) {
+        if ($this->isNewEntity($address) === true || $address->getParent()?->getId() !== $newParent->getId()) {
             $address->setParent($newParent);
             $address->setKey($this->getSaveKeyForMoving($address, $newParent));
 
--- a/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php
+++ b/src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php
@@ -126,7 +126,7 @@ final class CustomerTransformHelper implements CustomerTransformHelperInterface
         }
 
         // set new or changed parent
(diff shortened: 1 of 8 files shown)


src/CoreShop/Component/Core/Customer/CustomerTransformHelper.php
```diff
@@ -126,7 +126,7 @@ public function moveAddressToNewAddressStack(AddressInterface $address, ElementI
         }
 
         // set new or changed parent
-        if ($this->isNewEntity($address) === true || $address->getParent()->getId() !== $newParent->getId()) {
+        if ($this->isNewEntity($address) === true || ($address->getParent() && $address->getParent()->getId() !== $newParent->getId())) {
```
ramundomario, line 129:
```suggestion
        if ($this->isNewEntity($address) === true || $address->getParent()?->getId() !== $newParent->getId()) {
```
We could simplify with a null safe operator, WDYT? :

valentin-: @ramundomario looks good to me, updated.

ramundomario:
PSALM tests failing because of this:
https://github.com/coreshop/CoreShop/blob/19f36e427e424f6bc5523d604c545c0ea9c0361e/src/CoreShop/Component/Resource/Pimcore/Model/PimcoreModelInterface.php#L53-L56
Parent could be null, so PHPDoc should be:
```php
    /**
     * @return ElementInterface|null
     */
```
WDYT @dpfaffenbauer?

dpfaffenbauer:
@ramundomario yes

valentin-:
@ramundomario unfornuately the test still fails https://github.com/coreshop/CoreShop/runs/4656476471?check_suite_focus=true#step:14:42

dpfaffenbauer:
@valentin- That is not your issue now. I will fix the tests and you have to rebase this PR after that.

dpfaffenbauer:
@valentin- can you rebase now pls?

valentin-:
@dpfaffenbauer looks good now.

dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @valentin-

PR #1839, opened 2022-01-17
[Payment] fix for PayPal payment

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #...

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Migrations/Version20220117153741.php
+++ b/src/CoreShop/Bundle/CoreBundle/Migrations/Version20220117153741.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace CoreShop\Bundle\CoreBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20220117153741 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if (!$schema->getTable('coreshop_payment')->hasColumn('details')) {
+            $this->addSql("ALTER TABLE coreshop_payment ADD details LONGTEXT NOT NULL COMMENT '(DC2Type:json_array)';");
+        }
+    }
+
+    public function down(Schema $schema): void
+    {
+
+    }
+}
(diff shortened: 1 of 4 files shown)


src/CoreShop/Bundle/PaymentBundle/Resources/config/doctrine/model/Payment.orm.xml
```diff
@@ -12,6 +12,7 @@
         <field name="state" column="state"/>
         <field name="orderId" column="orderId" type="integer"/>
         <field name="datePayment" column="datePayment" type="datetime"/>
+        <field name="details" column="details" type="json_array" />
```
jdreesen, line 15: Please be aware that this type is [deprecated since DBAL 2.6](https://www.doctrine-project.org/projects/doctrine-dbal/en/2.13/reference/types.html#json-array) and was removed in DBAL 3 (`json` should be used instead).

As far as I can see, this is use first usage of the `json_array` type in CoreShop.
If so, it's probably better to directly use `json` instead.
If not, maybe all other usages could be converted before the final release of CoreShop 3, so it's possible to use it together with DBAL 3 (here's a blog post about [upgrading `json_array` to `json`](https://dunglas.fr/2022/01/json-columns-and-doctrine-dbal-3-upgrade/)).

dpfaffenbauer: thanks, changed

src/CoreShop/Bundle/CoreBundle/Migrations/Version20220117153741.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        if (!$schema->getTable('coreshop_payment')->hasColumn('details')) {
+            $this->addSql("ALTER TABLE coreshop_payment ADD details LONGTEXT NOT NULL COMMENT '(DC2Type:json_array)';");
```
jdreesen, line 20: Here should be `json` instead of `json_array` too, no?

jdreesen: See #1840

dpfaffenbauer: thanks for the PR, that is a beer on me 🍻  ;)

PR #1840, opened 2022-01-18
Fix doctrine type in migration (followup of #1839)

jdreesen:
I think this should be `json` instead of `json_array`, too.


dpfaffenbauer:
obviously... thanks

PR #1842, opened 2022-01-19
Fix rounding on tax calculation

podlom:
| Q                      | A
| -------------   | ---
| Bug fix?            | yes
| New feature?   | no
| BC breaks?      | no
| Deprecations? | no

For 3 items with € 16.99, the total amount must be € 50.97 and not € 50.98 which is wrong because it does not round correctly.
Then we have 19% tax and the calculation is for the net prices.
It fixes incorrect tax calculations by more accurate tax amount rounding.
The new correct calculation formula is:
taxAmount = round(net price * 1.19, 2) * quantity
In the current implementation calculation was:
taxAmount = round(net price * 1.19) * quantity


dpfaffenbauer:
@podlom can you please sign the CLA and fix the tests?

podlom:
> sign the CLA

I have signed the CLA.
How can I fix the failed tests?

dpfaffenbauer:
https://github.com/coreshop/CoreShop/pull/1849 fixes this as well. Please check there how we are going to fix it.

PR #1841, opened 2022-01-18
Slow load / save performance of objects due to serialized data

BlackbitNeueMedien:
We had the problem that saving and loading products took much longe than Pimcore's default objects. Then we wondered why loading objects was faster when we did not use object caching. The reason was the unserialization of the cached object because there was a lot of unnecessary data in the cache item `object_1234`: 
```
O:45:"AppBundle\Model\DataObject\ArticleDbExtension":147:{
...
some other field data
...
[...]

PR #1843, opened 2022-01-19
[Cache] optimize Pimcore cache with doctrine entities

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no

Related to https://github.com/coreshop/CoreShop/pull/1841


jdreesen:
You mean: related to #1841

src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/PimcoreCachePass.php
```diff
@@ -0,0 +1,31 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+
+final class PimcoreCachePass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $container->getDefinition('pimcore.cache.adapter.pdo')->setArgument(4, []);
+        $container->getDefinition('pimcore.cache.adapter.pdo')->setArgument(5, new Reference(CacheResourceMarshaller::class));
+
+        $container->getDefinition('pimcore.cache.adapter.redis_tag_aware')->setArgument(4, CacheResourceMarshaller::class);
```
jdreesen, line 29: Shouldn't this be a reference too?

```suggestion
        $container->getDefinition('pimcore.cache.adapter.redis_tag_aware')->setArgument(4, new Reference(CacheResourceMarshaller::class));
```

dpfaffenbauer:
> You mean: related to #1841

obviously :D

dpfaffenbauer:
@jdreesen what do you think of the solution? actually that should be somehting pimcore does, but they need to refactor serialization in general and this is something we cannot wait for.

src/CoreShop/Bundle/ResourceBundle/Pimcore/CacheResourceMarshaller.php
```diff
@@ -0,0 +1,107 @@
...
+    protected EntityManagerInterface $entityManager;
+
+    public function __construct(EntityManagerInterface $entityManager, MarshallerInterface $defaultMarshaller = null)
+    {
+        if (null === $defaultMarshaller) {
+            $defaultMarshaller = new DefaultMarshaller();
+        }
+
+        $this->defaultMarshaller = $defaultMarshaller;
+        $this->entityManager = $entityManager;
```
jdreesen, line 37:
```suggestion
        $this->defaultMarshaller = $defaultMarshaller ?? new DefaultMarshaller();
        $this->entityManager = $entityManager;
```

src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/PimcoreCachePass.php
dpfaffenbauer: yes thanks, didn't test redis yet.

src/CoreShop/Bundle/ResourceBundle/Pimcore/CacheResourceMarshaller.php
dpfaffenbauer: way better, thanks

src/CoreShop/Bundle/ResourceBundle/Pimcore/CacheResourceMarshaller.php
```diff
@@ -0,0 +1,103 @@
...
+                    }
+
+                    if (!$fd instanceof Data) {
+                        continue;
+                    }
+
+                    $data->setObjectVar(
+                        $fd->getName(),
+                        $fd->marshalForCache($data, $data->getObjectVar($fd->getName()))
+                    );
```
jdreesen, line 62: `$data` is the already deep-copied version of the original object, right? Otherwise, we'd override the properties and the original object wouldn't be usable anymore.

jdreesen (commented):
> what do you think of the solution?

I'm not really experienced in Pimcore's/Symfony's cache implementation, but at first glance it looks good to me.

> actually that should be somehting pimcore does, but they need to refactor serialization in general

:100: :+1:

src/CoreShop/Bundle/ResourceBundle/Pimcore/CacheResourceMarshaller.php
dpfaffenbauer: There is no deep-copy for cache. The idea here is to marshal the data per type and store it into the Object as anything you want. This only works as long as the properties are not fully typed. In my test case, I replace the `TaxRule` with the `id` of the `TaxRule`. The unmarshal then does the reverse. 

It is basically the same as the DeepCopy marshaller things work, but without the heavy overhead of DeepCopy.

BlackbitNeueMedien:
Can this solution be adjusted to make it compatible with Pimcore 6.9 / Coreshop 2.2?

dpfaffenbauer:
@BlackbitNeueMedien not sure really... Pimcore 6.9 still uses it's own cache implementation and I don't think that there are extensions points in there.

BlackbitNeueMedien:
Still struggling with performance issues. Is it correct that for example in https://github.com/coreshop/CoreShop/blob/bb08732d5a36e33f59663e925e6e2d8fe087a588/src/CoreShop/Bundle/ProductBundle/CoreExtension/ProductSpecificPriceRules.php#L202-L218 the object gets saved with JMS serializer? If yes, is it correct that in https://github.com/coreshop/CoreShop/tree/master/src/CoreShop/Bundle/CoreBundle/Resources/config/serializer there is no serialization rule for `productSpecificPriceRules`?

dpfaffenbauer:
@BlackbitNeueMedien  you get this issues with 2.2 right?

dpfaffenbauer:
Marshalling Data for Caching works similar as for Versioning. Both is not done wright. But it is how it is... You can try doing the same for Versioning. But: Reverting Versions might result in unexpected behaviour....

PR #1847, opened 2022-01-21
[IndexBundle] Migrate to SettingsStoreAwareInstaller

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

This PR implements the `SettingsStoreAwareInstaller` for the `CoreShopIndexBundle`.

### DIFF
--- a/src/CoreShop/Bundle/IndexBundle/Installer.php
+++ b/src/CoreShop/Bundle/IndexBundle/Installer.php
@@ -51,17 +51,11 @@ class Installer extends SettingsStoreAwareInstaller
         parent::install();
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function needsReloadAfterInstall(): bool
     {
         return true;
     }
 
-    /**
-     * {@inheritdoc}
-     */
     public function getOutput(): OutputInterface
     {
         return new NullOutput();
--- a/src/CoreShop/Bundle/IndexBundle/Installer.php
+++ b/src/CoreShop/Bundle/IndexBundle/Installer.php
@@ -15,16 +15,17 @@ declare(strict_types=1);
 namespace CoreShop\Bundle\IndexBundle;
 
 use Pimcore\Console\Application;
-use Pimcore\Extension\Bundle\Installer\InstallerInterface;
+use Pimcore\Extension\Bundle\Installer\SettingsStoreAwareInstaller;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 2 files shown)


src/CoreShop/Bundle/IndexBundle/Installer.php
```diff
@@ -46,33 +47,21 @@ public function install(): void
...
 
+    /**
+     * {@inheritdoc}
+     */
     public function needsReloadAfterInstall(): bool
     {
         return true;
     }
 
+    /**
```
dpfaffenbauer, line 62: remove doc blocks

src/CoreShop/Bundle/IndexBundle/Installer.php
```diff
@@ -15,16 +15,17 @@
...
 use Pimcore\Console\Application;
-use Pimcore\Extension\Bundle\Installer\InstallerInterface;
+use Pimcore\Extension\Bundle\Installer\SettingsStoreAwareInstaller;
 use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 
-class Installer implements InstallerInterface
+class Installer extends SettingsStoreAwareInstaller
```
dpfaffenbauer, line 24: what about the migration? bundle should be marked installed with a migration

dpfaffenbauer (changes requested):

src/CoreShop/Bundle/IndexBundle/Installer.php
solverat: Why? Just do a `parent::install()` in your `install()` method which will execute the initial setting-store stuff.

dpfaffenbauer: I am talking about already installed IndexBundles. With this change, these would not be marked as installed anymore

dpfaffenbauer: True... I think we can then ignore it

aarongerig: Sorry, just to make it clear: By "ignore it" you mean, remove the migration? 😅

dpfaffenbauer: @aarongerig Since you added it, we can keep it now :) thanks for that

dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @aarongerig

PR #1854, opened 2022-02-03
Added payum payment model fields to coreshop payment model

BabovicT:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

With Payment model not longer extending Payum/Payment model there is issue with orm mapping. Not all columns are being created in database. Also payment is not working because some fields couldn't be mapped because of that. 
This is proposition for first hand solution so that when someone installs CoreShop 3 payment works.

### DIFF
--- a/src/CoreShop/Bundle/PaymentBundle/Resources/config/doctrine/model/Payment.orm.xml
+++ b/src/CoreShop/Bundle/PaymentBundle/Resources/config/doctrine/model/Payment.orm.xml
@@ -20,6 +20,18 @@
             <gedmo:timestampable on="update"/>
         </field>
 
+        <field name="number" column="number" type="string" nullable="true" />
+
+        <field name="description" column="description" type="string" nullable="true" />
+
+        <field name="clientEmail" column="client_email" type="string" nullable="true" />
+
+        <field name="clientId" column="client_id" type="string" nullable="true" />
+
+        <field name="totalAmount" column="total_amount" type="integer" nullable="true" />
+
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
@BabovicT what is failing cause of missing fields?

BabovicT:
@dpfaffenbauer because of totalAmount not being in database amount paid is always 0. If we use for example paypal, it will be charged correct amount from it but in administration order would still be in state "awaiting payment". Same thing if we try to add payment through admin.
Other columns are not connected to that but some data is not being saved to database because it was missing. 
As I already wrote, this is just so fresh install can process payments. We are currently trying CoreShop 3(we have prihect in Pimcore X) so far thus was only thing missing for it to function like it was before.

BabovicT:
image
This is from Your demo site. 

image
As you can see there is no totalAmount column so every payment is always displayed as 0.

dpfaffenbauer:
ok, interesting. I tried avoiding these fields since we don't need them. I will look into the root cause of this issue

BabovicT:
There was already totalAmount, currencyCode and details fields in payment.php so i tought you just forgot to add them in orm. Glad that at least i pointed to this problem. You can close this pr then.

dpfaffenbauer:
I'll keep it open for now, maybe this is the solution anyway. But I'll check it

dpfaffenbauer:
@BabovicT I changed the PR. CientId and ClientEmail is never used. Other than that everything is perfect

dpfaffenbauer:
@BabovicT thanks

PR #1855, opened 2022-02-07
Password reset: Clearer form labels

BlackbitNeueMedien:
Currently there is a bit of a UX problem when resetting the password. We had some customers who wondered why they have to type their password when they actually want to reset it. To make it clearer that with the password-reset form it is actually possible to change to a **new** password, this PR changes the field labels.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1872, opened 2022-02-22
FIX: localized fields from brick and fieldcollections can be added to index

BabovicT:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | 

Brick and fieldcollections with localazied fields are now indexable.

### DIFF
--- a/src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
+++ b/src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
@@ -16,10 +16,16 @@ namespace CoreShop\Component\Index\Getter;
 
 use CoreShop\Component\Index\Model\IndexableInterface;
 use CoreShop\Component\Index\Model\IndexColumnInterface;
+use CoreShop\Component\Resource\Translation\Provider\TranslationLocaleProviderInterface;
 use Pimcore\Model\DataObject\Fieldcollection;
+use Pimcore\Model\DataObject\Localizedfield;
 
 class FieldCollectionGetter implements GetterInterface
 {
+    public function __construct(protected TranslationLocaleProviderInterface $localeProvider)
+    {
+    }
+
     public function get(IndexableInterface $object, IndexColumnInterface $config): mixed
     {
         $columnConfig = $config->getConfiguration();
@@ -38,15 +44,22 @@ class FieldCollectionGetter implements GetterInterface
                  */
                 $className = 'Pimcore\Model\DataObject\Fieldcollection\Data\\' . $columnConfig['className'];
                 if (is_a($item, $className)) {
+
                     $validItems[] = $item;
                 }
             }
         }
 
         foreach ($validItems as $item) {
+            $fallbackMemory = Localizedfield::getGetFallbackValues();
+            Localizedfield::setGetFallbackValues(true);
             if (method_exists($item, $fieldGetter)) {
-                $fieldValues[] = $item->$fieldGetter();
+                foreach ($this->localeProvider->getDefinedLocalesCodes() as $locale) {
+                    $fieldValues[$locale] =  $item->$fieldGetter($locale);
+                }
             }
+
+            Localizedfield::setGetFallbackValues($fallbackMemory);
         }
 
         return count($fieldValues) > 0 ? $fieldValues : null;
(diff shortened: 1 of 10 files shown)


src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
```diff
@@ -45,7 +51,19 @@ public function get(IndexableInterface $object, IndexColumnInterface $config): m
 
         foreach ($validItems as $item) {
             if (method_exists($item, $fieldGetter)) {
-                $fieldValues[] = $item->$fieldGetter();
+                if (!empty((new \ReflectionMethod($item, $fieldGetter))->getParameters())) {
```
dpfaffenbauer, line 54: feels a bit dirty.... we might wanna rather use ClassDefinition\FieldCollection Definition to check if the field is localized or not.

src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
```diff
@@ -45,7 +51,19 @@ public function get(IndexableInterface $object, IndexColumnInterface $config): m
...
+
+                    $fallbackMemory = DataObject\Localizedfield::getGetFallbackValues();
```
dpfaffenbauer, line 56: CoreShop already has a Helper for inheritance to make that better readable and re-usable, we should introduce that for Localized Fallbacks as well. (see https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Component/Pimcore/DataObject/InheritanceHelper.php for more).

dpfaffenbauer (changes requested):

dpfaffenbauer:
psalm issue is fixed: https://github.com/coreshop/CoreShop/pull/1873, can you rebase?

BabovicT:
@dominikager I also added Helper for inheritance to localizedFieldGetter.

src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
BabovicT: didn't know about it. Tnx

src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
BabovicT: Hope this one is more cleaner.

dpfaffenbauer:
@BabovicT Changed to how I imagine it, can you check and test pls?

src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
```diff
@@ -44,9 +51,30 @@ public function get(IndexableInterface $object, IndexColumnInterface $config): m
...
+
+            if (!$fd) {
+                continue;
+            }
+
+            if (!method_exists($item, $fieldGetter)) {
+                continue;
             }
+
+            if ($fd instanceof Localizedfield) {
```
BabovicT, line 64: @dpfaffenbauer I think it would work but $fd is not instance of LocalizedField. It is instance of field under the localizedfields. Thats why I used
`                if ($item->getDefinition()->getFieldDefinition('localizedfields')) {
`

dpfaffenbauer: but that only checks if there are localized-fields in the fieldcollection... which is useless basically. We need to know if the selected field is localized or not. But I see that there is an issue here. I will check again.

BabovicT: For some reason that checks if that field is inside localized-fields. If you have input and localized input that will be true only for localized input. Not sure why is that the case tho.

dpfaffenbauer: changed some more, worked for me now.

src/CoreShop/Component/Index/Getter/FieldCollectionGetter.php
```diff
@@ -44,9 +52,31 @@ public function get(IndexableInterface $object, IndexColumnInterface $config): m
...
             }
+
+            if (!method_exists($item, $fieldGetter)) {
+                continue;
+            }
+
+            if ($localizedFd) {
+                LocaleFallbackHelper::useFallbackValues(function() use ($item, $fieldGetter, &$fieldValues) {
+                    foreach ($this->localeProvider->getDefinedLocalesCodes() as $locale) {
+                        $fieldValues[$locale][] = $item->$fieldGetter($locale);
```
BabovicT, line 73:
```suggestion
                        $fieldValues[$locale] = $item->$fieldGetter($locale);
```
It doesn't work with [] for me. Can you test it one more time?

BabovicT:
Working for me too.

dpfaffenbauer (approved):

dpfaffenbauer:
thanks @BabovicT

PR #1882, opened 2022-03-01
[ResourceBundle] Make stack of classes not extended in CoreBundle working

davidhoeck:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

The stack for `coreshop.manufacturer` and `coreshop.address` was not working, because the classes have not been extended in the CoreBundle. Therefore I adjusted the check in the StackClassesPass to also work with not extended classes.


davidhoeck:
@dpfaffenbauer should be done

dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1883, opened 2022-03-02
FIX: coreshop_payment_provider editable select, wrong property name

BabovicT:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

coreshop_payment_provider editable was using property Named 'name' instead of the 'title'.


dpfaffenbauer:
thanks

PR #1896, opened 2022-03-10
Support class names whose name begins with lowercase letter

BlackbitNeueMedien:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Steps to reproduce bug:
1. Create class with lowercase first letter, e.g. product, extend `\CoreShop\Component\Core\Model\Product`
2. Set up the custom class configuration like described in https://docs.coreshop.org/2.2.0/Development/Extending_Guide/Extend_CoreShop_DataObjects.html:
    ```yaml
    core_shop_product:
        pimcore:
            product:
                classes:
                    model: 'Pimcore\Model\DataObject\Product'
3. Go to "cart price rules", add new rule, add `Products` condition, click "Search" -> you see that the defined Product class is not preselected.

The other fixed thing is when you override a data object class and set `core_shop_product.pimcore.product.classes.model` to the overridden FQCN (e.g. \App\Model\DataObject\Product), then  https://github.com/coreshop/CoreShop/blob/aff8053fcd566bc1462112550cda413c854013e8/src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/StackClassesPass.php#L58 does not have any effect.


dpfaffenbauer:
@BlackbitNeueMedien I fixed tests for 2.2, can you rebase to latest 2.2?

BlackbitNeueMedien:
Have I missed something? I merged up-to-date 2.2 branch but it complains about missing database columns.

src/CoreShop/Bundle/ResourceBundle/DependencyInjection/Compiler/StackClassesPass.php
```diff
@@ -55,11 +57,12 @@ public function process(ContainerBuilder $container)
                 if (in_array($interface, class_implements($definition['classes']['interface']))) {
                     $classStack[$alias][] = $definition['classes']['model'];
 
-                    $fullClassName = $definition['classes']['model'];
-                    $class = str_replace('Pimcore\\Model\\DataObject\\', '', $fullClassName);
-                    $class = str_replace('\\', '', $class);
-
-                    $classStackPimcoreClassName[$alias][] = $class;
+                    try {
+                        $reflectionClass = new ReflectionClass($definition['classes']['model']);
```
dpfaffenbauer, line 61: @BlackbitNeueMedien  this can be tricky, the class could not exist yet, since they either weren't installed yet or not created yet.

dpfaffenbauer (changes requested):

BlackbitDevs:
The problem still exists, also in 3.0 - do you know why you closed it, @dpfaffenbauer ? Because https://github.com/coreshop/CoreShop/pull/1896#pullrequestreview-982714728 can be easily fixed.

dpfaffenbauer:
This is actually quite difficult to solve. Problem is that during Container build time the ClassDefinitions cannot be loaded cause some of them require the kernel to be booted. So to fix this, I load the tokens from the definition file and try to find the classname, feels a bit dirty to me honestly, you might wanna check: https://github.com/coreshop/CoreShop/pull/2097

PR #1900, opened 2022-03-14
FIX: extending condition proccesor adds empty tag

BabovicT:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

Extending filter condition processor adds empty tag to definition. It probably isn't imagined to be extended but it should cover that case also.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1911, opened 2022-03-16
Fix xvfb error

dturopoli:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

The comment in code suggests that XVFB is optional, so getter for XVFB shouldn't throw exception.
`
 // use xvfb if possible
`

Also process needs to be created in a different way.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1914, opened 2022-03-18
fix: use only first-level categories in CategoryController::menuAction

dkarlovi:
Else all the categories are used, dumping them all in the main menu by default.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A

By default, category controller menu action will use all the available store categories to render the main menu. This doesn't work as expected since the main menu is rendered directly as if all the categories are root categories.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1941, opened 2022-03-29
Profiler fix

dturopoli:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

In case there is no store defined inside constructor code will enter catch part and $this->data will never be set, so the method getCurrency will throw and error and profile wont work


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #1979, opened 2022-05-02
Handle `null` addresses when persisting customers

jdreesen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When the customer that should be persisted in `CustomerManager::persistCustomer()` has no addessses yet, the property may be `null`, which fails when trying to iterate over it. We should fall back to an empty array in this case.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

Issue #2010, opened 2022-06-03
\CoreShop\Component\SEO\Extractor\ImageExtractor hardcodes Tool::getHostUrl() as prefix


dkarlovi:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

The thumbnail URL can be pointing to an external storage (via CDN) which is [configured](https://pimcore.com/docs/pimcore/10.0/Development_Documentation/Installation_and_Upgrade/System_Setup_and_Hosting/File_Storage_Setup.html) via
```yaml
pimcore:
   assets:
       frontend_prefixes:
           thumbnail: https://tavi-12345678990.cloudfront.net/thumbnail
```

dpfaffenbauer:
@dkarlovi https://github.com/coreshop/CoreShop/pull/2022

PR #2022, opened 2022-06-21
[SEO] Image Extractor should use frontend_prefix if configured

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2010


dpfaffenbauer:
tests failing for reasons it cannot download a package from gitlab, PR works though.

Issue #2023, opened 2022-06-21
Slow "object_url_slugs" queries on big project


hethehe:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

**Problem:** 
If you have a project with multiple sites (e.g. 4) and multiple languages (e.g. 6), then coreshop generates 30 URL slugs for every product ( (every site + the fallback) * every language).
On a project with 5000 products there are 150'000 entries.
=> The slugs are generated automatically on PRE_UPDATE and PRE_ADD

When you open a product in the backend, the URL slugs are queried:
`SELECT * FROM object_url_slugs WHERE fieldname = 'slug' AND ownertype = 'localizedfield' AND position = 'de' AND objectId = '1234'`
This query slows down with every new site, language, product.

**Possible Solutions:**
- Add an index for fieldname, ownertype, position and objectId
- Just add URL slugs for sites defined in the store section of coreshop
- Add an option "do not create slugs automatically for each site - use fallback for each site that does not have its own URL slug
- any more ideas?

dpfaffenbauer:
The slug field comes from Pimcore directly, so performance issues need to be reported there: https://github.com/pimcore/pimcore.

Regarding slugs only for valid sites: absolutely, would you create a PR for that?

PR #2025, opened 2022-06-21
create URL slugs only for valid coreshop sites/stores

hethehe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2023

CoreShop automatically creates the URL slugs for the products.
With this PR only slugs for available sites/store are considered. 

I have not changed the rest of the logic, that means other entries will be deleted.
Is this intentional?

To note:
In the URL slug definition of Pimcore there is an option "Available Sites" which is ignored in this PR as well in the current CoreShop version.

### DIFF
--- a/src/CoreShop/Bundle/PimcoreBundle/EventListener/SluggableListener.php
+++ b/src/CoreShop/Bundle/PimcoreBundle/EventListener/SluggableListener.php
@@ -17,9 +17,9 @@ namespace CoreShop\Bundle\PimcoreBundle\EventListener;
 use CoreShop\Component\Pimcore\Exception\SlugNotPossibleException;
 use CoreShop\Component\Pimcore\Slug\SluggableInterface;
 use CoreShop\Component\Pimcore\Slug\SluggableSluggerInterface;
-use CoreShop\Component\Store\Repository\StoreRepositoryInterface;
 use Pimcore\Event\DataObjectEvents;
 use Pimcore\Event\Model\DataObjectEvent;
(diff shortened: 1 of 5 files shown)


dpfaffenbauer:
@hethehe Please revisit that. The PimcoreBundle has no idea about the StoreBundle. The PimcoreBundle\SluggableListener is intended to be used "Standalone" as well. So you rather have to create a separate Listener that listens for "StoreAwareInterface" and creates the Slugs based on that.

hethehe:
Thanks for the clarification.

@dpfaffenbauer Another question/idea:
Is there a special reason we set the slug for each site and not just the fallback, does Coreshop need this explicity?

If not, I would change my PR so that the listener only sets the fallback and does not delete any entries.
Because actually the slug is set the same for each site, which is the same behavior as a fallback (but would save a lot of database entries).

dpfaffenbauer:
Well, good question... but honestly, 150k entries should not make things slower, pimcore needs to add indices.

hethehe:
I have opened a PR on Pimcore for the performance issue: https://github.com/pimcore/pimcore/issues/12486

But I still like the idea of not creating unnecessary DB entries :)
I have updated my PR so that only the fallback UrlSlug is created/updated automatically and only if the slug is empty.
=> Currently, all slugs are overwritten each time the object is saved.

I also had to update the SluggableLinkGenerator because it was not yet aware of the fallback slug.

PR #2030, opened 2022-06-27
Wishlist

IvanJavorovic:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

WIP new Wishlist bundle added to replace existing wishlist functionality. Review and changes are needed.


dkarlovi:
Looking at the PR, it's very nice progress. I think the direction is good, but since the change set is so large, it seems to me there's a lot of similarities with Cart that we should be able to reuse more code maybe?

For example, let's say we introduce a "product comparison" feature in the future, it would also be a list of products which you can add/remove to a separate storage, it would seem wasteful to need to again add so much functionality. With this in mind, is there anything we could do to abstract code for better reuse here?

dpfaffenbauer:
@dkarlovi Absolutely! The plan was to first create the wishlist and check what we need, and then further refactor and abstract it. Stuff could move into the StorageList Component and maybe we also have to create a StorageListBundle to abstract and combine stuff. But I am looking into that

dpfaffenbauer:
@dkarlovi reworked a lot now to share more between Order and Wishlist, you might wanna check?

dpfaffenbauer (approved):

dpfaffenbauer:
thanks @IvanJavorovic for kicking this off

PR #2032, opened 2022-06-30
[FrontendBundle] Consider UrlSlugs in the locale switcher

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

This PR enables the `coreshop_locale_switcher` Twig helper function to consider UrlSlugs (e.g. Products and Categories) and thus makes it possible for the user to switch the language and stay on the current product/category.

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
+++ b/src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
@@ -99,11 +99,12 @@ final class LocaleSwitcherExtension extends AbstractExtension
 
     private function getMainRequest(): Request
     {
-        $masterRequest = $this->requestStack->getMainRequest();
-        if (null === $masterRequest) {
+        $mainRequest = $this->requestStack->getMainRequest();
+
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
```diff
@@ -74,4 +96,14 @@ public function getLocalizedLinks(Document $document): array
...
     }
+
+    private function getMainRequest(): Request
+    {
+        $masterRequest = $this->requestStack->getMainRequest();
+        if (null === $masterRequest) {
+            throw new \UnexpectedValueException('There are not any requests on request stack');
+        }
+
+        return $masterRequest;
```
jdreesen, line 107:
```suggestion
        $mainRequest = $this->requestStack->getMainRequest();
        if (null === $mainRequest) {
            throw new \UnexpectedValueException('There are not any requests on request stack');
        }

        return $mainRequest;
```

PR #2038, opened 2022-07-08
fix: process conditions after the paginator

dkarlovi:
Reordering this allows for an optimization in some cases.

For example, when using an ES search here, you'll want to do the query and load the possible filter values in the same query. The query needs limit/offset for the items, but doesn't need it for filter values. Making the paginator run first allows us to inject the current limit/offset in the query early and we don't need to run it twice.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | N/A


dkarlovi:
Nevermind, the paginator needs the filters already applied. Need to find a different way to have limit/offset set before the paginator runs. :thinking:

PR #2055, opened 2022-08-02
[Reports] Support filtering for order type

BlackbitNeueMedien:
This PR adds an order type filter to the products, category and manufacturer reports because for reporting it is sometimes necessary to show all orders, not only the completed ones.

Additionally with this PR the category report supports maintaining product-category relation on category side (many-to-many object relation in `Category` class with allowed class `Product`).
And the manufacturer report now also supports the `name` field to be localized.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no


dpfaffenbauer:
@BlackbitNeueMedien I think it would be good if we use all available states and not statically implement them in Frontend. States are adjustable and can be altered per Project.

You already have the states in ExtJs as DataStore, you just either need to add the "all" field or allow the Filter to be empty to get all states. The Store is accessible with `pimcore.globalmanager.get('coreshop_states_order')`

BlackbitNeueMedien:
Thanks. Have applied the suggested changes. I removed the `all` option because setting no value means *all items* everywhere in Pimcore, so it can be the same here.

dpfaffenbauer (approved):

dpfaffenbauer:
thanks @BlackbitNeueMedien

PR #2058, opened 2022-08-19
[VariantBundle] Serializer: Allow $innerObject to be null

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

With this PR the value of `$innerObject` is allowed to be null when serializing a variant group. This prevents throwing an error when rendering the product detail page and thus makes this process more flexible.


dpfaffenbauer:
why can it be null?

aarongerig:
Because one of my color choices is basically a fallback value with no actual color code.
Bildschirmfoto 2022-08-19 um 08 43 30

All of the values which do not map to an existing color attribute are assigned this "other / unknown" color value.
Bildschirmfoto 2022-08-19 um 08 43 49

These values are fed by an external API, over which I do not have any control. This way I'm able to categorize those "other" values to this fallback attribute, which would be really flexible IMO.

dpfaffenbauer:
legit

aarongerig:
Thank you!

Issue #2040, opened 2022-07-12
[Pimcore UI] Default Product Unit always persisted, cannot unselect


aarongerig:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | yes

If at least one product unit is defined and you open the tab «Price & Units» of a Product object in the Pimcore UI, it is impossible to save a it without the default product unit definition.

See the behavior in the attached video.

https://user-images.githubusercontent.com/17384333/178445389-12825ad1-5204-4330-a55e-9467dfa833c0.mov

PR #2065, opened 2022-09-06
[Pimcore UI] Make default Product Unit unselectable

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2040

I'm not 100% sure if this is the correct place to fix the issue of not having an empty entry in the store. @dpfaffenbauer Could you please check and advise if there's a "better way" of adding an empty entry to a CoreShop resource list.
Thanks!


dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @aarongerig

PR #2074, opened 2022-09-13
fix #2073 by locking pkg version to 4.0.*

### LINKED ISSUE #2073: jms/serializer-bundle breaking change
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | yes
| RFC?             | no

I have installed `coreshop/core-shop: 3.0.0-beta.5` and at the some point composer pkg `jms/serializer-bundle` was updated from `4.0.1` to `4.1.0` and that cause following error

image (23)

gjorgic:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2073 

Lock pkg version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,7 @@
     "doctrine/orm": "^2.9",
     "fakerphp/faker": "^1.16",
     "gedmo/doctrine-extensions": "^3.6",
-    "jms/serializer-bundle": "^4.0",
+    "jms/serializer-bundle": "4.0.*",
     "knplabs/knp-menu-bundle": "^3.1",
     "laminas/laminas-stdlib": "^3.6",
     "laminas/laminas-zendframework-bridge": "^1.5",
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,7 @@
     "doctrine/orm": "^2.9",
     "fakerphp/faker": "^1.16",
     "gedmo/doctrine-extensions": "^3.6",
-    "jms/serializer-bundle": "4.0.*",
+    "jms/serializer-bundle": "^4.0",
     "knplabs/knp-menu-bundle": "^3.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -125,6 +125,7 @@
     "vimeo/psalm": "^4.10"
   },
   "conflict": {
+    "jms/serializer-bundle": "4.1.0 || 4.1.1"
```
jdreesen, line 128: How do you know that `4.1.1` will be broken, too?

gjorgic: not at all, that was reason why i lock version on the first time but you give thumbs up for this approach

gjorgic: but honestly, how can we be sure that they will even remove the bug in either of those two versions

jdreesen: It won't be removed in `4.1.0` as it's already released, thus we know it's broken.

Yes, we can't be sure it'll be fixed in the next version, but as it's blocking usage in dev mode, there's a high chance it will be resolved.

gjorgic: So if bug is still available, we will update `conflict` with new one and so on?

dpfaffenbauer: @gjorgic yes. But I am certain it will be resolved in 4.1.1

dpfaffenbauer:
@gjorgic please also sign the CLA

gjorgic:
@dpfaffenbauer now i did, please check now

composer.json
gjorgic: ok, i'm asking in general how you will handle if something like that happen, thx

dpfaffenbauer (approved):

dpfaffenbauer:
thanks @gjorgic

PR #2095, opened 2022-10-21
fix: index table o_classId type fix

dturopoli:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Since pimcore o_classId is string (varchar(50)) and inside index table o_classId is set to int type, values are not saved correctly inside index table


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #2102, opened 2022-11-08
[CustomerAssignment] Fix typo when passing the company ID to the router

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Before this PR the following URL is generated:
```log
/admin/coreshop/customer-company-modifier/dispatch-existing-assignment/16685/undefined
```

Which leads to following error:
```log
CoreShop\Bundle\CoreBundle\Controller\CustomerTransformerController::dispatchExistingAssignmentAction():
Argument #3 ($companyId) must be of type int, string given
```

With this PR, the correct company ID gets passed to the router and there generates the correct URL.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks @aarongerig

PR #2103, opened 2022-11-08
add option 'update-classes' for InstallResourcesCommand

lukadschaak:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | -

Add command option `update-classes` to the `InstallResourcesCommand` (`coreshop:resources:install`) to have the possiblity to update Pimcore class definitions.

When the command is executed, configuration from a bundles dependency injection (`Configuration.php`) is used to read in exported json files and to update the definition files for Pimcore.

In the current process, field definitions and objectbrick definitions are always updated, but class definitions never. So i thought why not have an option for classes as well.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

Issue #2115, opened 2022-11-16
[CartPriceRules] - Error complete order with card price rules based on cart items


TobiSW:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Referring #2057 

Setup is like the following. 

![image](https://user-images.githubusercontent.com/6861689/202176819-6d85ad1b-081b-45ce-939c-bdd464129938.png)

![image](https://user-images.githubusercontent.com/6861689/202176923-f33d4cff-4909-4180-a60e-dc159ed70c12.png)

![image](https://user-images.githubusercontent.com/6861689/202177084-fb718805-a461-481d-bd9a-b31a684a95f3.png)

After filling out the checkout steps when i click on "buy" i receive this error

![image](https://user-images.githubusercontent.com/6861689/202177259-3d88480b-d2df-43a7-9843-249d482e64c4.png)

dkarlovi:
AFAIK this is fixed in CS 3.0.2

dpfaffenbauer:
yes, thanks @dkarlovi

PR #2116, opened 2022-11-18
[CartPiceRules] fix OrderItem not finding Order with CartItemPriceRules

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2115

PR #2117, opened 2022-11-18
[StorageListItem] add definitive StorageList (Order/Wishlist) field to Item

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2115 

@TobiSW This is a different approach to the issue. This solves the underlying issue of the relation between order and order-item (eg. storage-list and storage-list-item).

@dkarlovi @solverat @breakone  Any thoughts on that?

### DIFF
--- a/src/CoreShop/Component/Wishlist/Model/Wishlist.php
+++ b/src/CoreShop/Component/Wishlist/Model/Wishlist.php
@@ -35,6 +35,13 @@ abstract class Wishlist extends AbstractPimcoreModel implements WishlistInterfac
     {
         Assert::isInstanceOf($item, WishlistItemInterface::class);
 
+        /**
+         * @var WishlistItemInterface $item
+         */
+        Assert::isInstanceOf($item, WishlistItemInterface::class);
+
+        $item->setWishlist($this);
+
         $items = $this->getItems();
         $items[] = $item;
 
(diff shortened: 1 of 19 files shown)


solverat:
Mhmm. It would be interesting to know why the message appears at all. Is it because in V2 the objects were saved more often?

However, it's a relation more but also no more do/while magic.

dpfaffenbauer:
@solverat It happens because the order gets moved and processed before saved. so the items are still in the old location.

solverat:
Ah I see, that's because of the cart/order behavior (same object)!

Well, that's going to be a heavy migration then. :)

dpfaffenbauer:
@solverat exactly. About the migration: Well, yes... There is the other PR https://github.com/coreshop/CoreShop/pull/2116 too, which just does a save before doing the calculations. 

But I think in the long term, it is better to have a real relation between the two.

src/CoreShop/Component/Wishlist/Model/Wishlist.php
```diff
@@ -35,6 +35,13 @@ public function addItem($item): void
     {
         Assert::isInstanceOf($item, WishlistItemInterface::class);
 
+        /**
+         * @var WishlistItemInterface $item
+         */
+        Assert::isInstanceOf($item, WishlistItemInterface::class);
```
jdreesen, line 41: Now the check is duplicated.

dpfaffenbauer: thanks

PR #2148, opened 2022-12-02
Add layout `price_rules` only once to OrderItem

m0nken:
When deciding if the new layout `price_rules` and field `priceRuleItems` need to be added to the class definition use the field's name for existence check and not the layout's name. Otherwise its existence is not properly detected which leads to multiple inserts if the migration is called multiple times or the field had already been added otherwise.

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2148


dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @m0nken

PR #2164, opened 2022-12-22
re-implement payum payment interface

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

This PR re-implements the `\Payum\Core\Model\PaymentInterface`. Otherwise, the Payum Actions won't work anymore. Example:

https://github.com/karser/PayumSaferpay/blob/master/Action/ConvertPaymentAction.php#L57

### DIFF
--- a/src/CoreShop/Component/Payment/Model/Payment.php
+++ b/src/CoreShop/Component/Payment/Model/Payment.php
@@ -24,7 +24,7 @@ use CoreShop\Component\Resource\Model\TimestampableTrait;
 /**
  * @psalm-suppress MissingConstructor
  */
-class Payment extends \Payum\Core\Model\Payment implements PaymentInterface
+class Payment implements PaymentInterface
 {
     use SetValuesTrait;
     use TimestampableTrait;
@@ -40,9 +40,21 @@ class Payment extends \Payum\Core\Model\Payment implements PaymentInterface
     protected $paymentProvider;
 
     /**
+     * @var int
+     */
+    protected $totalAmount;
+
+    /**
+     * @var string
+     */
+    protected $currencyCode;
+
+    /**
      * @var string
      */
     protected $state = PaymentInterface::STATE_NEW;
+
+    protected array $details = [];
 
     /**
      * @var \DateTime
@@ -53,6 +65,16 @@ class Payment extends \Payum\Core\Model\Payment implements PaymentInterface
      * @var int
      */
     protected $orderId;
+
+    /**
+     * @var string
+     */
+    protected $number;
+
+    /**
+     * @var string
+     */
+    protected $description;
 
     public function getId()
     {
@@ -67,6 +89,26 @@ class Payment extends \Payum\Core\Model\Payment implements PaymentInterface
     public function setPaymentProvider(PaymentProviderInterface $paymentProvider)
     {
         $this->paymentProvider = $paymentProvider;
+    }
+
+    public function getTotalAmount()
+    {
+        return $this->totalAmount;
+    }
+
+    public function setTotalAmount($amount)
+    {
+        $this->totalAmount = $amount;
+    }
+
+    public function getCurrencyCode()
+    {
+        return $this->currencyCode;
+    }
+
+    public function setCurrencyCode($currencyCode)
+    {
+        $this->currencyCode = $currencyCode;
     }
 
     public function getDatePayment()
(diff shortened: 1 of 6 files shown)


src/CoreShop/Component/Payment/Model/Payment.php
```diff
@@ -24,7 +24,7 @@
 /**
  * @psalm-suppress MissingConstructor
  */
-class Payment implements PaymentInterface
+class Payment extends \Payum\Core\Model\Payment implements PaymentInterface
```
dpfaffenbauer, line 27: needs to be done in Core\Model\Payment, Payment Component does not have a dependency to Payum.

solverat:
Ok, I think there are more things to discuss, @dpfaffenbauer. 

CoreShop creates a valid `PayumPayment` object here:

https://github.com/coreshop/CoreShop/blob/c4ddac02c7df4a94f4ee363912ddc9a1322913f3/src/CoreShop/Bundle/PayumBundle/Action/AuthorizePaymentAction.php#L53

https://github.com/coreshop/CoreShop/blob/c4ddac02c7df4a94f4ee363912ddc9a1322913f3/src/CoreShop/Bundle/PayumBundle/Action/CapturePaymentAction.php#L55

However, there is another issue if we're going to keep it that way. I'm using your unzer gateway to show you the problem:

1. This action will fail because in the first try, it is not a payum model: https://github.com/coreshop/payum-unzer/blob/a00e9a441221ed3f48ee8213b177703e18c2a405/Action/ConvertPaymentAction.php#L52
2. Now the `catch()` method is going to execute the convert action again, this time with an payum model
3. But now, the extension will fail here, because it's not an coreshop payment interface anymore: https://github.com/coreshop/PayumUnzerBundle/blob/339f99d7bb3630f2edbc33f1b030828d04f18198/Extension/PopulateUnzerExtension.php#L50
 
My suggestion:
- Remove the try / catch and use the payum model instantly 
- Instead of the raw payum model, we could create `Core\Model\CoreShopPayumPayment` with an `$order` property

WDYT?

dpfaffenbauer:
@solverat here is what you can do with how CoreShop currently is designed:

- Package 1: Payum Unzer: No Dependency to CoreShop
- Package 2: Payum Unzer Bundle: CoreShop Integration of the Payment Provider

So as was before.

Then, in the Bundle, you create a `ConvertCoreShopPaymentAction` that looks something like this:

```php
final class ConvertCoreShopPaymentAction implements ActionInterface
{
    public function execute($request)
    {
        RequestNotSupportedException::assertSupports($this, $request);

        /** @var CoreShopPaymentInterface $payment */
        $payment = $request->getSource();
    }

    public function supports($request)
    {
        return
            $request instanceof Convert &&
            $request->getSource() instanceof CoreShopPaymentInterface &&
            $request->getTo() == 'array';
    }
}
```

and add it as service:

```yaml
    CoreShop\Payum\UnzerBundle\Action\ConvertCoreShopPaymentAction:
        tags:
            - { name: payum.action, alias: unzer_convert_action, factory: unzer, gateway: unzer, prepend: true }
```

In there, you have complete Control over the CoreShop Order and can access anything and prepare data for the Payment Provider.

solverat:
Hey @dpfaffenbauer:

"So as it was before": Almost. In CS2, the Payum Gateway also was able to manage the request source, because it was a real Payum Payment Interface.

Within your proposal, the CoreShop bundle always needs to copy the content of the Gatway's `ConvertPaymentAction`, because it will never apply - something I don't want to deal with. Every CoreShop Payum Bundle needs to adopt those changes. Sometimes there can be huge, if you're looking at existing extensions:

https://github.com/search?l=PHP&q=%22class+ConvertPaymentAction%22&type=Code

So I'm far away from accepting this as a solution. But I can't do more than suggest an alternative solution (personally, I would go with "Way II").

dpfaffenbauer:
@solverat It's too late for that now, CoreShop 3.0 is out and thats the way it is now.

solverat:
BTW: One of our devs (@Zodiarc) has found a much cleaner solution (And keep the extension at it is):

```php
// ConvertPaymentPageExtension
$paymentEntity = $this->paymentRepository->createQueryBuilder('p')
    ->where('p.number = :orderNumber')
    ->setParameter('orderNumber', $payment->getNumber())
    ->getQuery()
    ->getSingleResult();

if (!$paymentEntity instanceof \Coreshop\Component\Core\Model\Payment) {
    return;
}

$order = $paymentEntity->getOrder();

// ... append data to $request 
```

PR #2169, opened 2023-01-04
Prevent error when adding new store without currency

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

When a data object class uses field type `Store Values`, the store may not have a currency set yet.

Steps to reproduce bug:
1. Create product
2. Create new store (it gets created without currency - required field check is only done afterwards when editing existing store
3. Open product from 1.
You will get the error `Uncaught Error: Call to a member function getSymbol() on null`

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/CoreExtension/StoreValues.php
+++ b/src/CoreShop/Bundle/CoreBundle/CoreExtension/StoreValues.php
@@ -624,10 +624,12 @@ class StoreValues extends Model\DataObject\ClassDefinition\Data implements
                 continue;
             }
 
+            $currency = $store->getCurrency();
+
             //Fill missing stores with empty values
             $storeData[$store->getId()] = [
                 'name' => $store->getName(),
-                'currencySymbol' => $store->getCurrency()->getSymbol(),
+                'currencySymbol' => $currency?->getSymbol() ?? '',
                 'values' => ['price' => 0],
                 'inheritable' => $inheritable,
             ];


src/CoreShop/Bundle/CoreBundle/CoreExtension/StoreValues.php
```diff
@@ -624,10 +624,12 @@ public function getDataForEditmode($data, $object = null, $params = [])
                 continue;
             }
 
+            $currency = $store->getCurrency();
+
             //Fill missing stores with empty values
             $storeData[$store->getId()] = [
                 'name' => $store->getName(),
-                'currencySymbol' => $store->getCurrency()->getSymbol(),
+                'currencySymbol' => $currency ? $currency->getSymbol() : '',
```
jdreesen, line 632: We're on PHP 8, right? So why not:

```suggestion
                'currencySymbol' => $currency?->getSymbol() ?? '',
```

dpfaffenbauer:
@BlackbitDevs please rebase to 3.x

BlackbitDevs:
Rebased to 3.0.
@dpfaffenbauer What is actually the correct base branch for bugfixes / new features? 3.0 for bugfixes and master for new features?

dpfaffenbauer:
yes. 3.0 is for smaller features/enhancements and master (as 3.1) for new featuers

dpfaffenbauer (approved):

dpfaffenbauer:
thanks @BlackbitDevs

PR #2173, opened 2023-01-10
allow unpublished objects in index command

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

Unpublished objects are not available when executing the `coreshop:index` command. In some scenarios, this could lead to some orphans in the index (Error while removing an unpublished object from index will never ever get removed once an exception has been thrown)

Use `coreshop:index -u` to assert all objects are in sync, since it will include unpublished objects!


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #2201, opened 2023-02-03
[FrontendBundle] fix only showing customer orders and not cars too

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no

### DIFF
--- a/src/CoreShop/Component/Order/Repository/OrderRepositoryInterface.php
+++ b/src/CoreShop/Component/Order/Repository/OrderRepositoryInterface.php
@@ -39,6 +39,8 @@ interface OrderRepositoryInterface extends PimcoreRepositoryInterface, CustomerA
 
     public function findByCustomer(CustomerInterface $customer): array;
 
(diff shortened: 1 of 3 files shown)


src/CoreShop/Component/Order/Repository/OrderRepositoryInterface.php
```diff
@@ -39,6 +39,8 @@ public function findExpiredCarts(int $days, bool $anonymous, bool $customer): ar
 
     public function findByCustomer(CustomerInterface $customer): array;
 
+    public function findOrdersByCustomer(CustomerInterface $customer): array;
```
jdreesen, line 42: Strictly speaking, this is a breaking change, right?

dpfaffenbauer: Stricly speaking yes, but since this is sort of a serious bug, it's fine for us. There is no other way of solving this than to introduce a new Interface for a feature that is expected to work

PR #2209, opened 2023-02-10
Throw exception if site store isnt found in SiteBasedRequestResolver

pascalheidmann:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   

If you have a custom resolver with a priority lower than SiteBasedRequestResolver and are inside a request flagged as site request it will never come so far as it will return `null` instead of erroring


dpfaffenbauer:
I am not really sure about potential BC breaks, maybe others suspect that the default store is returned when there is no store for the site

pascalheidmann:
@dpfaffenbauer returning the default store would be also a valid solution for me (which errors if none found). But at least it should not fail silently/break the resolver which expects a valid store and not `null`

dpfaffenbauer:
@pascalheidmann I think we should do it. At least within 3.x

dpfaffenbauer (approved):

PR #2210, opened 2023-02-11
Require at least php 8 via composer.json

pascalheidmann:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Coreshop uses PHP8 language features and therefore should require it via `composer.json`.
This also has the nice benefit that at least in PHPStorm the language level is automatically pushed up to 8.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -84,6 +84,7 @@
     "coreshop/messenger-bundle": "self.version"
   },
   "require": {
+    "php": "^8.0",
     "ext-json": "*",
     "dachcom-digital/emailizr": "^2.0.0",
     "doctrine/data-fixtures": "^1.5",


dpfaffenbauer:
Pimcore requires PHP 8 by default, thats why we didn't define it specifically https://github.com/pimcore/pimcore/blob/10.6/composer.json#L31

jdreesen:
But it doesn't hurt, and it helps PhpStorm and maybe people just looking into this repo, right?

dpfaffenbauer:
Ok.

dpfaffenbauer:
thanks @pascalheidmann

Issue #2198, opened 2023-02-02
Only set object slug if not set


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

https://github.com/coreshop/CoreShop/blob/3b0eccb3d1f427ce4a9526a4ae3588796dc6459d/src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php#L81

If the slug has been set once, it never should get touched again from the `DataObjectSlugGenerator`. Currently, the slug gets updated and re-adjusted, if the category/product name has changed.

There are two reasons, why this should be done:
- SEO: If the content manager changes the product name, the URL should not (necessarily) change
- The content manager may want to change the slug to a complete different slug

TL;DR: Setting the slug automatically is ok - but only if the (localized) slug field is empty.

dpfaffenbauer:
SEO: Pimcore automatically creates a redirect for changed slugs, so the URL is still available.

solverat:
Yes, I'm aware of that. But every redirect counts as a negative impact and should be prevented as much as possible.

solverat:
@dpfaffenbauer Yes!! 💪

PR #2236, opened 2023-03-22
Make URL slugs overwritable

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no

Currently the URL slugs are always regenerated during saving. There is no way to manually change a URL. 
Even worse, in https://github.com/coreshop/CoreShop/blob/75d7f0b2e8d72b9537c948d896c8e6d2e27bd356/src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php#L67-L68
`$newSlug` is the generated one and `$existingSlug` is the one which may have been changed with the objects setter method (`$sluggable->setSlug()`). 
Example:
I have a product with `name=my product` and set the slug in the pimcore backend to be `/en/product-123`. When this gets saved, the slug `/en/my-product` gets generated. With above code not only the manually set slug gets reset to `/en/my-product` but moreover there will also be a redirect generated for `/en/product-123` to `/en/my-product`.

With this PR the URL slug gets generated if none exists but it does not use the generated one when there already is a slug set.

### DIFF
--- a/src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php
+++ b/src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php
@@ -70,8 +70,10 @@ class DataObjectSlugGenerator implements DataObjectSlugGeneratorInterface
                             // $existingSlug is the slug to be saved from backend
                             $dbSlug = null;
                             if($sluggable instanceof Concrete) {
+                                /** @psalm-suppress InternalMethod */
                                 $dbSlug = $sluggable->retrieveSlugData(['fieldname' => 'slug', 'ownertype' => 'object', 'position' => $language, 'siteId' => $existingSlug->getSiteId()])[0]['slug'] ?? null;
                                 if ($dbSlug === null) {
+                                    /** @psalm-suppress InternalMethod */
                                     $dbSlug = $sluggable->retrieveSlugData(['fieldname' => 'slug', 'ownertype' => 'object', 'position' => $language])[0]['slug'] ?? null; // fallback slug
                                 }
                             }


src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php
```diff
@@ -41,7 +42,10 @@ public function generateSlugs(SluggableInterface $sluggable): void
                 new UrlSlug($fallbackSlug, 0),
             ];
             $actualSlugs = [];
+            $inheritanceEnabled = AbstractObject::getGetInheritedValues();
```
dpfaffenbauer, line 45: use CoreShops Inheritance Helper

dpfaffenbauer (changes requested):

src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php
BlackbitDevs: Thx, done.

BlackbitDevs:
@dpfaffenbauer What do you think about PhpStan complaining about `Call to an undefined method CoreShop\Component\Pimcore\Slug\SluggableInterface::retrieveSlugData()`? This is a Pimcore core method https://github.com/pimcore/pimcore/blob/765832f0dc5f6cfb296a82e089b701066f27bcef/models/DataObject/Concrete.php#L849 - add it to the `SluggableInterface`?

solverat:
Related issue: https://github.com/coreshop/CoreShop/issues/2198

src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php
```diff
@@ -41,7 +43,9 @@ public function generateSlugs(SluggableInterface $sluggable): void
                 new UrlSlug($fallbackSlug, 0),
             ];
             $actualSlugs = [];
-            $existingSlugs = $sluggable->getSlug($language);
+            $existingSlugs = InheritanceHelper::useInheritedValues(function() use ($sluggable, $language) {
+                return $sluggable->getSlug($language);
+            }, false);
```
jdreesen, line 48:
```suggestion
            $existingSlugs = InheritanceHelper::useInheritedValues(
                fn () => $sluggable->getSlug($language),
                false,
            );
```

dpfaffenbauer:
@BlackbitDevs adding the function to the SluggableInterface also means that "parent" classes need to implement it. What if we type check for Concrete?

BlackbitDevs:
> @BlackbitDevs adding the function to the SluggableInterface also means that "parent" classes need to implement it. What if we type check for Concrete?

Perfectly fine for me ;)
Will adjust the code.

dpfaffenbauer:
@BlackbitDevs can you check the tests pls?

BlackbitDevs:
Ok, the remaining Psalm error is
`InternalMethod - src/CoreShop/Component/Pimcore/Slug/DataObjectSlugGenerator.php:73:55 - The method Pimcore\Model\DataObject\Concrete::retrieveSlugData is internal to Pimcore but called from CoreShop\Component\Pimcore\Slug\DataObjectSlugGenerator::generateSlugs`
What to do? Copy the method to Coreshop? Or ask Pimcore to remov the `@internal`?

dpfaffenbauer:
@BlackbitDevs  ```/** @psalm-suppress InternalMethod */```

dpfaffenbauer:
thanks a lot @BlackbitDevs

PR #2237, opened 2023-03-22
re-implement expression evaluation

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

This PR enables the expression language,

### DIFF
--- a/src/CoreShop/Bundle/ResourceBundle/Validator/Constraints/UniqueEntityValidator.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Validator/Constraints/UniqueEntityValidator.php
@@ -39,7 +39,7 @@ final class UniqueEntityValidator extends ConstraintValidator
     protected function evaluateExpression(array $value): array
     {
         foreach ($value as $i => $item) {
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
Shouldn't we revert that to how it was before before?

```php
    protected function evaluateExpression($value)
    {
        if (is_array($value)) {
            foreach ($value as $i => $item) {
                $value[$i] = $this->evaluateExpression($item);
            }

            return $value;
        }

        try {
            return $this->expressionLanguage->evaluate($value, ['container' => $this->container,]);
        } catch (SyntaxError $e) {
            // do not throw any exception but simple return value instead
            return $value;
        }
    }
```

solverat:
@dpfaffenbauer I also thought about this. 

The reason this has been changed is that the validator now only accept an `array` in `fields`, therefore it is not required anymore to check it against the `is_array`statement.

But that's just a guess.

dpfaffenbauer:
I see @solverat , than its fine

dpfaffenbauer (approved):

PR #2240, opened 2023-03-27
Fix multiselect filter when work with array

lexwm:
There was a problem with the display of the multiselect filter on the frontend.
During the analysis, it turned out that the multiselect processor extracts all possible options for a specific field and then works with the array.
The current kernel code did not assume this.

I've made corrections to imply that the selected filter value can be not only a string or a number, but it can also be an array.


dpfaffenbauer:
@lexwm Please also sign the CLA

PR #2252, opened 2023-04-15
[Next] Pimcore 11 compatibility

dpfaffenbauer:
Keeping compatibility between Pimcore 10 and 11, is not worth the effort. CoreShop 3.1 will be the last feature release, 4.0 will then only be compatible with Pimcore 11.

Todo:
 - [ ] Workflow Callbacks - dependency inject services that are used
 - [x] Admin Controllers: Move to Service Subscribers


dpfaffenbauer:
FINALLY ALL GREEN :)

PR #2254, opened 2023-04-18
Use shipping address of logged in customer for tax calculation

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no

Assume we have a store with base country DE:
Bildschirmfoto 2023-04-18 um 14 02 26

And we have tax rule 19% VAT for country Germany and 0% VAT for all other countries:
Bildschirmfoto 2023-04-18 um 14 02 05

Initially without being logged-in the taxes shown on the website will be based on the store's base country. This is correct.
But when the client logs in, we already know his default address. In the checkout this gets used anyway but not when browsing through the site. On the site we still see taxes based on the store's base country. 
With this PR the logged-in customer's address will get used for displaying taxes also when browsing the shop.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks @BlackbitDevs

PR #2259, opened 2023-04-25
Update docs

dpfaffenbauer:


solverat:
image

🤯

Nice!

dpfaffenbauer:
@solverat I think most of it comes from using the same Theme as Pimcore :D. But I also adapted the Configuration a lot.

PR #2261, opened 2023-04-26
Adjusted locale switcher for internal shop pages without translations

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Existing locale switcher doesn't change locale for static route pages such as /shop/login, /shop/cart, /shop/customer/profile etc. This happens because `$document` in https://github.com/coreshop/CoreShop/blob/738d977fea986e9ab13145c53b25ee00ed84f454/src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php#L50 relates to the document whose leftmost path matches the current URL. When you only have `/shop` as document but the remaining CoreShop pages are handled via static routes, even a request for `/shop/login` will have `$document = /shop`. Of course this document may have translations and so the wrong links get created.

This PR checks if a static route has been used for current request. In this case links for this static route (but for other locale) will get created. This avoids having to add all internal shop pages as Pimcore Documents.

### DIFF
--- a/src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
+++ b/src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
@@ -92,7 +92,7 @@ final class LocaleSwitcherExtension extends AbstractExtension
                 $route = $this->getMainRequest()->attributes->get('_route');
                 $staticRoute = Staticroute::getByName($route);
                 $params = [];
-                if (strpos($staticRoute->getVariables(), '_locale')) {
+                if ( str_contains($staticRoute->getVariables(), '_locale') ) {
                     $params = ['_locale' => $language];
                 }
                 $link = $this->container->get('router')->generate($route, $params);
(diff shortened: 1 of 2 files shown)


src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
```diff
@@ -35,6 +37,7 @@ public function __construct(
         private Document\Service $documentService,
         private ShopperContextInterface $shopperContext,
         private RequestStack $requestStack,
+        protected ContainerInterface $container,
```
dpfaffenbauer, line 40: never ever inject the whole container! inject what you need, in this case RouterInterface.

dpfaffenbauer (changes requested):

src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
dpfaffenbauer: @BlackbitDevs I made the change now. We are preparing the release today

dpfaffenbauer (approved):

src/CoreShop/Bundle/FrontendBundle/Twig/LocaleSwitcherExtension.php
BlackbitDevs: Thanks @dpfaffenbauer

PR #2262, opened 2023-04-27
Support "store" field type in grid

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

If show a `coreShopStore` (store select) field in grid you'll get no data

Снимок экрана 2023-04-27 в 17 02 21

This is because `CoreShop\Bundle\StoreBundle\CoreExtension\Store` class hasn't method `getDataForGrid`

This method with all dependent methods `setOptions` and `getOptions` was copied from https://github.com/pimcore/pimcore/blob/10.5/models/DataObject/ClassDefinition/Data/Select.php (because Coreshop does not extend Pimcore's Select field type)
Only return data was changed from `return $data;` to `return $data?->getId();` because the getter returns a `Store` object via https://github.com/coreshop/CoreShop/blob/bb483e7279ac28af236caf8e8ff96001661cf70f/src/CoreShop/Bundle/ResourceBundle/CoreExtension/Select.php#L186

The most important point why it did not work is that in
https://github.com/pimcore/pimcore/blob/a5b86f3f9a63ac76282acee00879fb52b416a4ce/bundles/AdminBundle/Resources/public/js/pimcore/object/tags/select.js#L99
`field.layout.optionsProviderClass` is checked. This refers to a public field `optionsProviderClass` in the Store PHP class which did not exist before (only the `getOptionsProviderClass()` existed) and caused a JS error because of accessing a non-existing object property.

With this PR it looks like this
Снимок экрана 2023-04-27 в 17 04 10

### DIFF
--- a/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
+++ b/src/CoreShop/Bundle/StoreBundle/CoreExtension/Store.php
@@ -35,7 +35,7 @@ class Store extends Select
      */
     public $fieldtype = 'coreShopStore';
 
-    /** @var array */
+    /** @var array|null */
     public $options = [];
 
     /** @var string */
@@ -63,7 +63,7 @@ class Store extends Select
 
     public function getOptionsProviderClass()
     {
-        return '@' . StoreOptionProvider::class;
+        return $this->optionsProviderClass;
     }
 
     public function getDataForGrid($data, $object = null, $params = [])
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
@BlackbitDevs NICE :), can we rebase it to 3.0?

BlackbitDevs:
Is now rebased to 3.0

dpfaffenbauer:
@BlackbitDevs tests are failing

dpfaffenbauer (approved):

dpfaffenbauer:
@BlackbitDevs thank you very much

PR #2270, opened 2023-05-04
rename Behaviour to Behavior 

nosovk:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

We use cspell in our projects, and found that project has some words mistyped. In this PR we renamed all `Behaviour` to `Behavior` in codebase


jdreesen:
Well, this is a breaking change...

nosovk:
yep, found [that](https://www.scribbr.co.uk/uk-vs-us/behavior-or-behaviour/#:~:text=In%20UK%20English%2C%20'behaviour',')%20is%20the%20correct%20spelling.) also. Then will just correct CSpell dictionary.

dpfaffenbauer:
@nosovk I will not merge that, there is no benefit in having this. I use british english a lot since I work with british people a lot. Thanks for your contribution though.

PR #2279, opened 2023-05-11
[ClassDefinitionPatchBundle] introduce new Bundle to allow patching Pimcore Class Definitions

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no

This feature allows to patch class definitions and preview the patches for Pimcore Class Definitions. Use Case is mainly for other Extensions that need new fields or changes on Class Definitions.

Example:

```yaml
core_shop_class_definition_patch:
    patches:
        '%coreshop.model.company.pimcore_class_name%':
            fields:
                name2:
                    before: 'addresses'
                    definition:
                        fieldtype: input
                        title: coreshop.company.name2
        '%coreshop.model.customer.pimcore_class_name%':
            fields:
                name2:
                    before: 'firstname'
                    definition:
                        fieldtype: input
                        title: coreshop.company.name2

```

patch


src/CoreShop/Bundle/ClassDefinitionPatchBundle/Console/ColorConsoleDiffFormatter.php
```diff
@@ -0,0 +1,112 @@
...
+
+            return $string;
+        }, $escapedDiffLines);
+
+        return sprintf($template, implode(PHP_EOL, $coloredLines));
+    }
+
+    private function makePlusLinesGreen(string $string): string
+    {
+        return Strings::replace($string, self::PLUS_START_REGEX, '<fg=green>$1</fg=green>');
```
jdreesen, line 100: Is there a reason you're using the Nette tools here, instead of the [Symfony string component](https://symfony.com/doc/current/components/string.html#methods-to-search-and-replace)?

dpfaffenbauer: I copied that part from somewhere else. I dind't wanna include a package just for this. I can check if Symfony String also solves this. Makes one less dependency.

dpfaffenbauer: @jdreesen works with Symfony String, thanks for the hint 💪

PR #2282, opened 2023-05-11
support 'zero' values for filtering (#2278)

### LINKED ISSUE #2278: Support "zero" filter values
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

If a (boolean) column is in the index with possible values `0` or `1`, then it's impossible to add a filter condition
matching all `0`s. In the field where the value(s) can be selected, only `1` is displayed.

The reason for this is this check for a "truthy" value:
https://github.com/coreshop/CoreShop/blob/master/src/CoreShop/Component/Index/Worker/FilterGroupHelper.php#L83

This check should either be removed or adjusted e.g. `if (null === $e)`

benwalch:
this PR fixes #2278

### DIFF
--- a/src/CoreShop/Component/Index/Filter/BooleanFilterConditionProcessor.php
+++ b/src/CoreShop/Component/Index/Filter/BooleanFilterConditionProcessor.php
@@ -62,7 +62,7 @@ class BooleanFilterConditionProcessor implements FilterConditionProcessorInterfa
             $value = $condition->getConfiguration()['preSelect'];
         }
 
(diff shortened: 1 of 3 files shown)


benwalch:
@dpfaffenbauer pls restart failed job since github was not reachable

benwalch:
@dpfaffenbauer thx - pls merge ;)

dpfaffenbauer:
Thanks a lot @benwalch

PR #2291, opened 2023-05-24
changed jmsSerializer type from ArrayTransformerInterface to Serializ…

BlackbitDevs:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

Adjusted OrderController according to JsonHelperTrait Pimcore using SerializerInterface instead of ArrayTransformerInterface
```
https://github.com/pimcore/pimcore/blob/81be9535e5b461fe045c0cce6cf3d94be92901bf/bundles/AdminBundle/Controller/AdminController.php#L34:36
```
```
abstract class AdminController extends UserAwareController implements AdminControllerInterface
{
    use JsonHelperTrait;
```


dpfaffenbauer:
@BlackbitDevs that doesn't work, CoreShop is using JMS Serializer, where as Pimcore injects the Symfony Serializer. We already changed that in the Pimcore 10.6 compatibility PR: https://github.com/coreshop/CoreShop/pull/2285

PR #2294, opened 2023-05-26
improve unlinking files in pdf generation

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

1. `$file` can be null, it is not allowed to pass `null`in `unlink`
2. Use `file_exists` instead of `@unlink` (see https://php.watch/versions/8.0/fatal-error-suppression)


dpfaffenbauer (approved):

dpfaffenbauer:
thanks

PR #2295, opened 2023-05-26
[SerializedData] show data as string  in version preview

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

image

This one is quite old. Until now we always had to override the pimcore admin quicksearch object section, otherwise the exeption above will raise up when a cart (or now an order) object gets transformed for the version preview. However, besides us, it seems that no one else is using the pimcore quicksearch :)

This PR simply retturns again the serialized data:

image


dpfaffenbauer (approved):

dpfaffenbauer:
@solverat never had that issue :D, thanks for the fix

PR #2313, opened 2023-06-09
add docs for creating custom order number generator

### LINKED ISSUE #2312: How to overwrite order number generation
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | no
| BC Break report? | no
| RFC?             | no
| Type | Just a question |

Hi!

Background: i want to change the order number that is generated when summiting an order via checkout.

The only solution it found was to create a new symfony service and decorate the default number generator. An example is below.

My question is: Is there a better way of doing this? in this configuration you have to work with `decoration_priority` to either be called before the `prefix_suffix` generator, or after it.

If you choose `decoration_priority` to be `>= 0`, the order of called decoraters would be  `PrefixSuffix Generator > PaddingNumberGenerator > DefaultSequenceGenerator`
If you choose a `decoration_priority: -1`, the order of called decoraters would change to `PaddingNumberGenerator > PrefixSuffix Generator > DefaultSequenceGenerator`

Additional question: If this implementation seems okay for you guys, do you think it is worth to document it?

```yaml
  App\Order\PaddingNumberGenerator:
    decorates: coreshop.order.number_generator.default
    decoration_priority: 1
    arguments:
      - '@.inner'
      - 8 # order number padding length
```

```php
final class PaddingNumberGenerator implements NumberGeneratorInterface
{
    public function __construct(
        private NumberGeneratorInterface $numberGenerator,
        private int $paddingLength,
    ) {
    }

    public function generate(ResourceInterface $model): string
    {
        $number = $this->numberGenerator->generate($model);

        return str_pad($number, $this->paddingLength, '0', STR_PAD_LEFT);
    }
}

lukadschaak:
This will add docs on how to create a custom number generator.

Closes #2312

### DIFF
--- a/docs/03_Development/06_Order/10_NumberGenerator.md
+++ b/docs/03_Development/06_Order/10_NumberGenerator.md
@@ -1,0 +1,39 @@
+# CoreShop Order Number Generator
+
+When an order is completed, a number is generated for that order. This happens at the `CoreShop\Component\Core\Order\Committer\OrderCommitter`, which uses a `NumberGeneratorInterface` service.
+
+If the CoreBundle is installed, the `coreshop.order.number_generator.prefix_suffix` service is injected. It decorates the `coreshop.order.number_generator.default` and extends it with the feature to manage a prefix and suffix via the CoreShop settings.
+
+To create a custom number generator, a service must be created which also decorates the `coreshop.order.number_generator.default`. To set the order of the decorations, use `decoration_priority`.
+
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
please rebase onto 3.0. next is not ready yet

lukadschaak:
is rebased to 3.0

lukadschaak:
Thx :+1:

PR #2344, opened 2023-07-17
[Menu] making it easer to have more independent bundles with the same main-menu

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no


jdreesen:
With `v3.1.1` using the `coreshop/messenger-bundle` as standalone is again broken. There is a CoreShop icon with a "Messenger" sub-item, but nothing happens when clicking on it. Nothing in the browser console, either.

dpfaffenbauer:
😢 what Pimcore Version?

jdreesen:
Pimcore `v10.5.25`.

It seems that the `type` is wrong and doesn't match the one in the listener:
grafik
It expects `coreshop.messenger` but it is `coreshop.coreshop`.

Maybe we should check `item.id` or `item.name` instead?
grafik

dpfaffenbauer:
it should be coreshop.coreshop. We changed how menues are loaded. Could you PR it? We can make a 3.1.3 release quite fast to relaese that fix

jdreesen:
Here it is: #2393

PR #2373, opened 2023-09-18
"apply CacheResourceMarshaller to doctrine_dbal adapter, add cache marshaller to MoneyCurrency

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

This PR fixes a quite huge memory issue, when it comes to cache marshalling of doctrine entities. Without it, out of memory exceptions will occur randomly while unserializing cache entries.

Background: In our case, we're dealing with large currency entities (lot of countries as relation for example), so we had cache entries up to 4-5mb each.

This PR fixes two things ( + performance bonus):
- Add CacheResourceMarshaller to `pimcore.cache.adapter.doctrine_dbal` Service
- Add marshalling to `MoneyCurrency`


dpfaffenbauer (approved):

dpfaffenbauer:
@solverat thanks

PR #2418, opened 2023-10-27
fix: orders have already reserved their items, no double-check necessary

### LINKED ISSUE #2417: [Bug] Payment after revising does not work properly
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

In the revise action the `PaymentType` form is used and validated just as during the checkout process. However when revising, the order already has the `SaleState` "order" what means the items are already on hold. As of now the `CardStockAvailabilityValidator` performs its check again. When one has the last item available reserved in the order, this check will fail, leading to not being able to checkout.

rberneder:
Fixes #2417

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2417

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/Validator/Constraints/CartStockAvailabilityValidator.php
+++ b/src/CoreShop/Bundle/CoreBundle/Validator/Constraints/CartStockAvailabilityValidator.php
@@ -22,6 +22,7 @@ use CoreShop\Component\Core\Model\OrderInterface;
 use CoreShop\Component\Core\Model\OrderItemInterface;
 use CoreShop\Component\Inventory\Checker\AvailabilityCheckerInterface;
 use CoreShop\Component\Inventory\Model\StockableInterface;
+use CoreShop\Component\Order\OrderSaleStates;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 use Webmozart\Assert\Assert;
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
@rberneder the solution is not right. Since the Order has been created doesn't necessarily mean we don't want to use that Constraint Validator anymore. I much rather use different Validation Groups for the revise than to just ignore the Validator.

rberneder:
@dpfaffenbauer I
- agree that bypassing the validation is not correct.
- think the behavior of the validator is also not correct. The availability of a product reflects the sum of all items in stock minus the ones on hold (of which the passed order/item is part of).

I think the best of both worlds is to consider the items of the passed order as part of the on hold count during validation. WDYT?

dpfaffenbauer:
The Validator only checks for available stock, which is onHand-onHold. 

What do you mean with passed order?

dpfaffenbauer:
I think I see what you mean. But that still is not valid. We have a new feature in 3.2 that allows changing a confirmed order, that currently doesn't check for Stock on changing the order, but then your proposed solution what still not be right. I think the best solution for now is to not run the validation in order-revise.

dpfaffenbauer:
@rberneder I am closing this for now, feel free to further comment and we can discuss how to solve this.

Issue #2425, opened 2023-10-30
Error in the filter functionality for multiselects


hethehe:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

There is an error in the filter functionality if you use any filter with multiple values (multiselects)

Since Symfony 6 we can't get a non scalar value form $parameterBag->get().
Symfony 5 triggered following deprecation note:

> Retrieving a non-scalar value from "$parameterBag->get()" is deprecated, and will throw a "%s" exception in Symfony 6.0, use "%s::all($key)" instead.

It is used in Component/Index/Filter/*

I think it should be fixed by changing "$parameterBag->get()" to "$parameterBag->all()".
PR is on the way.

PR #2426, opened 2023-10-30
Fix error in the filter functionality for multiselects

hethehe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2425

This PR fix an error in the filter functionality if you use any filter with multiple values (multiselects)

Since Symfony 6 we can't get a non scalar value form $parameterBag->get().
Symfony 5 triggered following deprecation note:

> Retrieving a non-scalar value from "$parameterBag->get()" is deprecated, and will throw a "%s" exception in Symfony 6.0, use "%s::all($key)" instead.


dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @hethehe

Issue #2445, opened 2023-11-13
[Bug] NestedConditionChecker NOT validation broken


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Within [this huge commit](https://github.com/coreshop/CoreShop/commit/efbcf5197f699902e184cf083fd3ed17d5cb226e), the variable `$valid` changed to `$conditionValid`:

![image](https://github.com/coreshop/CoreShop/assets/700119/7f4ddedc-ed9a-4cc1-99bb-b3ea42b0b36d)

That means, that if you're using the `NOT` operator, `$valid` will always be `false` and therefor the rule always returns `true`.

dpfaffenbauer:
thanks a lot for the PR @solverat

PR #2447, opened 2023-11-14
fix nested rule condition

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2445 

I THINK this should fix the **not** condition in nested rules.


dpfaffenbauer:
@solverat thanks for the fix, I will add a test so we don't fall into a regression anymore

dpfaffenbauer:
@solverat test is added, it actually fixes it. without your Pr, the test fails

dpfaffenbauer (approved):

Issue #2367, opened 2023-08-24
Inconsistent Cart Item <> Cart Storage Behavior


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

https://github.com/coreshop/CoreShop/blob/836bb013a1163c6471aa8b288998a6483775bb60/src/CoreShop/Bundle/OrderBundle/Manager/CartManager.php#L74-L84

After all items have been persisted, the cart processor kicks in:

https://github.com/coreshop/CoreShop/blob/836bb013a1163c6471aa8b288998a6483775bb60/src/CoreShop/Bundle/OrderBundle/Manager/CartManager.php#L87

In some cases, the cart processor throws an exception. After that, the customer cannot add any items anymore, because of this:

```bash
Duplicate full path [ /coreshop/carts/2023/08/24/cart64e73dc0a5d463.86004598/items/1 ] - cannot save object
``` 

It's quite obvious why this is happening: The cart object does not hold any items in its relations (because the cart persistence crashed), but physically they exist.

Solution:
- Check the path via `DataObject\Service::pathExists()`
- OR: Create an unique key (1-xy)

dpfaffenbauer:
why does to cart processor fail? thats the main problem here

solverat:
Yes. We can't tell why it's happening at now, but there are a LOT of processes (all cart processors + a lot of post update events).

Sometimes it's even okay, if there is an exception. So, after the customer has been informed by any kind of error (and changed something - custom additional text for example), it is not possible to re-add an item ever again.

dpfaffenbauer:
welcome to a world were Pimcore doesn't allow transactions with dataobjects and concurrency ;)

I honestly don't know how to properly solve it. We could either try-catch the cart processors and save the cart afterwards to get the relations in order. but what side-effects does that come with?

solverat:
> welcome to a world were Pimcore doesn't allow transactions with dataobjects and concurrency ;)

Yes.... :angry: 

Speaking of transactions, :), the craziest thing we could do:

```php
$this->connection->beginTransaction();

try {

   // save/create all order items
   // save/process cart

    $this->connection->commit();

} catch (\Throwable $e) {

    if ($this->connection->isTransactionActive()) {
        $this->connection->rollBack();
    }

    throw $e;
}
```

[...]

PR #2510, opened 2024-01-16
[CartManager] create unique key for cart items

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2367 

@solverat WDYT of this solution? I think they key as index is not really necessary


solverat:
@dpfaffenbauer Sorry, totally missed that. I guess this will work, thanks!

PR #2520, opened 2024-01-25
[Core] add caching for recursive variant checking

### LINKED ISSUE #2519: findRecursiveVariantIdsForProductAndStoreByProducts Performance / Bug
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

## Performance
https://github.com/coreshop/CoreShop/blob/dd7aadeeeb5af284d9e47be47719d4db9a94bb94/src/CoreShop/Bundle/CoreBundle/Pimcore/Repository/ProductRepository.php#L64-L71

To demonstrate, here's a product which we need to display with all its possible variant prices in a table:

#### With enabled findRecursiveVariantIdsForProductAndStoreByProducts 
image

However, improving the query doesn't change a thing. The real problem is, that (depending on the project and price rules you're using) this query gets executed a million times. This leads to database crashes.

## Bug
I also noted, that this method does not respect products given store!

### Cache Entry
Instead of executing the same query over and over, we've added a short cache item:

```php
$cacheKey = sprintf('cc_rvips_%s_%d', md5(implode('-', $products)), $store->getId());
Cache::save($resultProducts, $cacheKey, [], 500, 0, true);
```

### Summary

```php
$cacheKey = sprintf('cc_rvips_%s_%d', md5(implode('-', $products)), $store->getId());

if(false !== $cacheData = Cache::load($cacheKey)) {
    $resultProducts = $cacheData;
} else {

    /** @psalm-suppress InternalMethod */
    $query = sprintf('
        SELECT oo_id as id FROM (
            SELECT CONCAT(o_path, o_key) as realFullPath FROM objects WHERE o_id IN (:products)
        ) as products
        INNER JOIN %s variants ON variants.o_path LIKE CONCAT(products.realFullPath, "/%%")
        WHERE variants.stores LIKE :store
    ', $dao->getTableName());

    $params = [
        'products' => $products,
        'store'    => '%,' . $store->getId() . ',%'
    ];

    $paramTypes = [
        'products' => Connection::PARAM_STR_ARRAY,
    ];

    $resultProducts = $this->connection->fetchAllAssociative($query, $params, $paramTypes);

    Cache::save($resultProducts, $cacheKey, [], 500, 0, true);
}

// [...]
```

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no

fixes #2519

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/EventListener/RecursiveVariantCacheInvalidationEventListener.php
+++ b/src/CoreShop/Bundle/CoreBundle/EventListener/RecursiveVariantCacheInvalidationEventListener.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * CoreShop
+ *
+ * This source file is available under two different licenses:
+ *  - GNU General Public License version 3 (GPLv3)
+ *  - CoreShop Commercial License (CCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ * @copyright  Copyright (c) CoreShop GmbH (https://www.coreshop.org)
+ * @license    https://www.coreshop.org/license     GPLv3 and CCL
+ *
+ */
+
+namespace CoreShop\Bundle\CoreBundle\EventListener;
+
+use CoreShop\Bundle\ResourceBundle\Event\ResourceControllerEvent;
+use CoreShop\Component\Core\Payment\Rule\Condition\ProductsConditionChecker as PaymentRuleProductsConditionChecker;
+use CoreShop\Component\Payment\Model\PaymentProviderRuleInterface;
+use CoreShop\Component\Rule\Model\RuleInterface;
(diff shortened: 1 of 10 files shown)


solverat:
@dpfaffenbauer That looks good to me! :+1:  I just stumbled over an existing clear listener:

https://github.com/coreshop/CoreShop/blob/dd7aadeeeb5af284d9e47be47719d4db9a94bb94/src/CoreShop/Bundle/ProductBundle/Resources/config/services.yml#L117-L122

Don't know if this should be consolidated or at least the new payment rule cleared should be added?

dpfaffenbauer:
@solverat I am honestly not sure where that tag is set

dpfaffenbauer:
@solverat just figured it out, that is not used anymore. I'll delete it

PR #2524, opened 2024-01-26
use null coalescing operator against maxUsagePerUser

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | --

- Use null coalescing operator against `maxUsagePerUser` which is not available after updating coreshop on existing rules
- Do not use elvis operator since it's not safe against availability


solverat:
@dpfaffenbauer I just saw that the fix on `VoucherConditionChecker.php` already has been applied, but only on v4. 

https://github.com/coreshop/CoreShop/commit/5baaa1630d9e7d359f40be40b46b28c0c909cb1c

dpfaffenbauer (approved):

dpfaffenbauer (approved):

PR #2536, opened 2024-02-07
[ResourceBundle] auto registration of pimcore models

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      |no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no

### DIFF
--- a/src/CoreShop/Component/Resource/Reflection/ClassReflection.php
+++ b/src/CoreShop/Component/Resource/Reflection/ClassReflection.php
@@ -25,11 +25,7 @@ final class ClassReflection
     public static function getResourcesByPaths(array $paths): iterable
     {
         foreach ($paths as $resourceDirectory) {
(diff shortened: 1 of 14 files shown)


src/CoreShop/Component/Resource/Reflection/ClassReflection.php
```diff
@@ -0,0 +1,78 @@
...
+final class ClassReflection
+{
+    public static function getResourcesByPaths(array $paths): iterable
+    {
+        foreach ($paths as $resourceDirectory) {
+            $resources = self::getResourcesByPath($resourceDirectory);
+
+            foreach ($resources as $className) {
+                yield $className;
+            }
```
jdreesen: Can't you do it like this?
```suggestion
            yield from self::getResourcesByPath($resourceDirectory);
```

Issue #2541, opened 2024-02-11
Cart/Order Token Missing / Not Unique


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Since #2511, the checkout requires a valid token, to complete the order. 

We're struggling with two issues here:

- The token is not unique per se: Back-Check if an order exists with generated token? I know, changes are low, but it's not unique
- Carts which have been created before ##2265, don't come with any token => 404 in checkout (It's quite old, but still valid). 

Possible Solution: https://github.com/coreshop/CoreShop/issues/2539#issuecomment-1937492184

Or migrate it:

```php
protected function execute(InputInterface $input, OutputInterface $output): int
{
    $listing = new CoreShopOrder\Listing();
    $listing->setUnpublished(true);
    $listing->addConditionParam('token = "" OR token IS NULL');
    $listing->addConditionParam(sprintf('saleState = "%s"', OrderSaleStates::STATE_CART ));

    foreach ($listing->getObjects() as $cart) {

        $token = $this->tokenGenerator->generate(10);

        $output->writeln(sprintf('<comment>Set missing token %s for cart %d', $token, $cart->getId()));

        $cart->setToken($token);
        $cart->save();
    }

    return Command::SUCCESS;
}
```

dpfaffenbauer:
@solverat a migration can be very heavy on the database. So I'd rather use the fallback approach, WDYT?

dpfaffenbauer:
I'd prefer a migration anyway ;).

solverat:
:D yes, probably a fallback would be smoother

dpfaffenbauer:
follow up ticket for uniqueness check: https://github.com/coreshop/CoreShop/issues/2546

PR #2543, opened 2024-02-11
[Frontend] create order-token if not yet exists

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2541


solverat:
@dpfaffenbauer I know, it's annoying, but what will happen, if a generated token is not unique? The user won't be able to finalize the checkout because the `save` method will throw an exception from where the frontend can't recover.

dpfaffenbauer:
I thought about that, but I consider that quite unlikely. We never had that before. We should definitely add a check, but for now I think it is safe enough

PR #2544, opened 2024-02-11
[ProductBundle] default return empty array instead of null in preGetData

breakone:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #...


dpfaffenbauer (approved):

dpfaffenbauer:
@breakone can also happen in v3 right?

Issue #2532, opened 2024-02-05
Control discountability per Specific Price Rule


EinShoppo:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | yes/no
| RFC?             | no

Hello @dpfaffenbauer,

We have a requirement from a customer and I would like to hear your opinion on this as it is also potentially an interesting feature for CoreShop.

The starting point is the following: Our customer builds his product prices using the “Specific Price Rules”. There are different price rules for each customer group and time-based prices.
There are also “Cart Price Rules” that allow, for example, a 10% discount per item.
Now this discount should not always be granted, since books in Germany, for example, are subject to fixed prices. We have therefore added a new attribute "discountable" to the product class, which we check for within the "Cart Price Rule" using the CartItemPriceRule condition.

All of this works wonderfully, but now our customer would like to be able to use the “Specific Price Rule” to control whether the price can be discounted via the Cart Rule, not globally on the product. The background is that, for example, customer group A can receive a discount, customer group B however not. For another item it could be the other way around and for a third item the discount should in principle be possible, but not for a price scheduled for the future because the price already represents a special price.

As you can see, the customer needs a high degree of flexibility. To achieve this, my idea is as follows:
* OrderItems are provided with a new flag attribute "discountable" (default: true)
* Create a new "Specific Price Rule Action" called "non-discountable" (since discountable products are the default)
* When this action is applied, the OrderItem attribute discountable is set to false
* There is a new CartItemPriceRule condition "discountable" for the Cart Price Rules, which checks the value of the discountable attribute of the OrderItem

In this way it is possible to control the discountability depending on a specific price rule. Somehow I'm unsure whether the idea with the non-discountable action is a good one, but I can't think of another solution. What do you think about it? Would this be interesting for the core?

Best regards

dpfaffenbauer:
@EinShoppo I love the idea. We have to be careful about priorities. What if a price rule action triggers before discountable is set?

EinShoppo:
@dpfaffenbauer Oh, you're probably right. We ourselves always use the specific price rules with "stop propagation", so only one of these rules is executed for each product. But we could run into a prioritization problem if you have multiple specific rules for the same product and without stopping propagation after a valid rule.
[...]

PR #2555, opened 2024-02-18
[Order] allow passing custom-attributes from price rules to order-item

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2532

@EinShoppo this is not ready yet, but the basic idea behind what we discussed. The idea is to allow passing custom attributes from pricing calculation to be stored in the order, thus using them to affect price calculation from cart-price-rules. (or basically anything cart-item-related). WDYT?

It is still missing parts like persistence of attributes or the cart-price-rule-condition to not allow non-discountable.


dpfaffenbauer:
@EinShoppo rest is now implemented as well

dpfaffenbauer:
@EinShoppo do you wanna give this a test?

ThatTrick:
Hi @dpfaffenbauer,

sounds great! One of my colleagues and i are looking at this today. I think you'll get feedback from us tomorrow at the latest.

EinShoppo:
@dpfaffenbauer: Sounds great!  One of my colleagues and I are looking at this today. I think you'll get feedback from us tomorrow at the latest.

EinShoppo:
@dpfaffenbauer We finished our tests and it works like a charm! Big cheers and many thanks :+1:
I really like the fact that the attribute is added to the cart price rules validation automatically, because i think that this attribute should win in any case and so it's good that you don't have to set a condition manually.

A possible improvement would be to also check for the presence of the not-discountable attribute in the product price rules, but this is probably technically more difficult to implement. So you can see it more as an idea for the future.

dpfaffenbauer:
@EinShoppo nice :). Ok, then we can merge this into the 4.1 branch. you then have to update your project to 4.1 to use that ok?

EinShoppo:
@dpfaffenbauer We planed to update to 4.0 before, so this shouldn't be a big problem for us. Do you already know when it will be released?

dpfaffenbauer:
@EinShoppo probably after Pimcore Inspire. Are you faster than that?

EinShoppo:
@dpfaffenbauer: I'm not sure, but if necessary we'll have to temporarily fork or patch the 4.0 version.

dpfaffenbauer:
@EinShoppo let's have a chat next week and discuss that, can you contact me directly?

Issue #2572, opened 2024-03-04
Wrong Order Subtotal / Total Calculation when using Adjustments


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

- Add adjustment to order item via `coreshop.cart_processor` and priority `590` (for example `10.-`)
- Check total price in cart ⇾ should be ok
- Complete order
- Check total price in order ⇾ not ok

First, I was not sure if it has something to with the new immutable stuff (which is a hefty bc-break for custom cart processors and was undocumented).

But then I ran into other issues, and so I debugged it again. Still not sure, but I think the `CartSubtotalProcessor` is not allowed to recalculate adjustments in cart ⇾ order processing workflow:

```php
if ($cart->isImmutable()) {
    return;
}
```

After I've changed that, it works fine. Also, not sure why we didn't recognize it earlier... most probably, because my adjustment is not a discount but a simple surcharge on item layer.

dpfaffenbauer:
I don't quite understand. You're saying that the `CartSubtotalProcessor` should also return when immutable?

solverat:
Exactly. 

Because it triggers the `$cart->recalculateAdjustmentsTotal()` and that shouldn't be done, if the cart has been transformed => but that's something you may have to confirm.

dpfaffenbauer:
can you show me how your custom processor looks?

solverat:
```php
# tagged via coreshop.cart_processor, priority: 590
<?php declare(strict_types=1);

use CoreShop\Component\Order\Factory\AdjustmentFactoryInterface;
use CoreShop\Component\Order\Model\OrderInterface;
use CoreShop\Component\Order\Model\OrderItemInterface;
use CoreShop\Component\Order\Processor\CartProcessorInterface;
use CoreShop\Component\Product\Model\ProductInterface;
use CoreShop\Component\Registry\ServiceRegistryInterface;

final class MyAdjustmentProcessor implements CartProcessorInterface
{
    public function __construct(
        protected ServiceRegistryInterface $registry,
        protected AdjustmentFactoryInterface $adjustmentFactory
    ) {
    }

    public function process(OrderInterface $cart): void
    {
        if ($cart->isImmutable()) {
            return;
        }

        /** @var OrderItemInterface $item */
        foreach ($cart->getItems() as $item) {

            $product = $item->getProduct();
            if (!$product instanceof ProductInterface) {
[...]

PR #2573, opened 2024-03-04
[Core] fix price calculation with immutables and adjustments

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2572

### DIFF
--- a/src/CoreShop/Component/Core/Order/Processor/CartItemsImmutableProcessor.php
+++ b/src/CoreShop/Component/Core/Order/Processor/CartItemsImmutableProcessor.php
@@ -68,6 +68,14 @@ class CartItemsImmutableProcessor implements CartProcessorInterface
                 $itemDiscount,
                 $context
             );
+
+            /**
+             * https://github.com/coreshop/CoreShop/issues/2572
+             *
+             * Since we are recalculating the items-total, we also need to respect the adjustments and
+             * re-add them to the item total as well.
+            */
+            $item->recalculateAdjustmentsTotal();
         }
     }
 }
--- a/src/CoreShop/Component/Core/Order/Processor/CartItemsImmutableProcessor.php
+++ b/src/CoreShop/Component/Core/Order/Processor/CartItemsImmutableProcessor.php
@@ -69,12 +69,12 @@ class CartItemsImmutableProcessor implements CartProcessorInterface
                 $context
             );
 
-            /**
+            /*
              * https://github.com/coreshop/CoreShop/issues/2572
              *
              * Since we are recalculating the items-total, we also need to respect the adjustments and
              * re-add them to the item total as well.
-            */
+             */
             $item->recalculateAdjustmentsTotal();
         }
     }
--- a/src/CoreShop/Component/Core/Order/Processor/CartSubtotalProcessor.php
+++ b/src/CoreShop/Component/Core/Order/Processor/CartSubtotalProcessor.php
@@ -26,33 +26,7 @@ final class CartSubtotalProcessor implements CartProcessorInterface
 {
     public function process(OrderInterface $cart): void
     {
-        $subtotalGross = 0;
-        $subtotalNet = 0;
-
-        /**
-         * @var OrderItemInterface $item
-         */
-        foreach ($cart->getItems() as $item) {
-            /*
-             * https://github.com/coreshop/CoreShop/issues/2572
-             *
-             * We have to use the subtotal here.
-             *
-             * The difference between $item->getTotal and $item->getSubtotal is,
-             * that the total includes all applied adjustments from cart rules.
-             *
-             * We do that since we need to calculate tax for cart rules and split
-             * the rule amount onto the items.
-             *
-             * We also need to handle cart rule values for returns and credit-memos.
-             */
-            $subtotalGross += $item->getSubtotal(true);
-            $subtotalNet += $item->getSubtotal(false);
-        }
-
-        $cart->setSubtotal($subtotalGross, true);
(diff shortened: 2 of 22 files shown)


src/CoreShop/Component/Core/Order/Processor/CartSubtotalProcessor.php
```diff
@@ -33,8 +33,21 @@ public function process(OrderInterface $cart): void
          * @var OrderItemInterface $item
          */
         foreach ($cart->getItems() as $item) {
-            $subtotalGross += $item->getTotal(true);
-            $subtotalNet += $item->getTotal(false);
+            /**
```
jdreesen: This is a comment and not a DocBlock, right?
```suggestion
            /*
```

src/CoreShop/Component/Core/Order/Processor/CartItemsImmutableProcessor.php
```diff
@@ -68,6 +68,14 @@ public function process(OrderInterface $cart): void
                 $itemDiscount,
                 $context
             );
+
+            /**
+             * https://github.com/coreshop/CoreShop/issues/2572
+             *
+             * Since we are recalculating the items-total, we also need to respect the adjustments and
+             * re-add them to the item total as well.
+            */
```
jdreesen:
```suggestion
            /*
             * https://github.com/coreshop/CoreShop/issues/2572
             *
             * Since we are recalculating the items-total, we also need to respect the adjustments and
             * re-add them to the item total as well.
             */
```

solverat:
@dpfaffenbauer thanks for the quick response. Unfortunately, this does not work when using additional adjustments on cart items. In other words: The total value of the order items are getting completely ignored now.

To reproduce: You may want to add a simple adjustment to an item (not cart) in your test to validate the total value. 

I'm not sure why this PR is not failing, but I guess it's because `CartRuleApplier` is adding an adjustment on item- AND cart layer:

https://github.com/coreshop/CoreShop/blob/1ca9f58651ebb314a56a942d5c7e1df5e89eb423/src/CoreShop/Component/Core/Cart/Rule/Applier/CartRuleApplier.php#L189-L190

https://github.com/coreshop/CoreShop/blob/1ca9f58651ebb314a56a942d5c7e1df5e89eb423/src/CoreShop/Component/Core/Cart/Rule/Applier/CartRuleApplier.php#L200-L202

dpfaffenbauer:
@solverat adding an adjustment to the item without the order is not supported and is something we cannot make work easily without breaking other things. as you pointed out, cart-rules are applied on cart-item and cart directly. and that is the reason it is applied on both.

dpfaffenbauer:
@solverat one solution could be that we set the adjustment on the cart itself as neutral and use them from the items. not sure what side-effects that can have.

dpfaffenbauer:
@solverat please check the latest changes

solverat:
@dpfaffenbauer This looks VERY promising! 

In my Test, I've
- added a 2% global cart discount rule
- added a global surcharge cart rule with fix amount
- added a custom adjustment (just as before, only on item layer but with an `isImmutable` check)

…and all calculations are correct now!

The only thing I don't understand in your PR:

https://github.com/coreshop/CoreShop/pull/2573/commits/9e742c8660ef099ca0662eae6c5f95a6d8198cbd#diff-ad51a1ab22ce6a03fe767ee679b9ccfd41a00b017606df746e1baf50b4c5f490R61-R72

I removed that part, and I wasn't able to spot the difference. Maybe a hint to test it would be great!

dpfaffenbauer:
@solverat If you have already have items with a total of 0, without that, it would lead to a division by 0 error.

I did some other changes as well, can you test them as well? Since we are now changing which discounts apply to what, we need to make sure to always have the right calculations.

solverat:
@dpfaffenbauer Just adopted your changes: Still valid!

dpfaffenbauer:
I will also test it in one of our projects first. It might have unconsidered side-effects

solverat:
I also think that this PR makes the neutral adjustment calculation more solid. :+1: 

It's quite clear now, what's happening when adding custom adjustments - especially if they have to be added to the order item subtotal to allow them to be discountable by cart rules!

dpfaffenbauer:
Yeah, so basically what this changes is:
 - Cart Price Rules are applied to the Cart Total, and added to the Cart Item as Neutral Adjustment
 - Cart Price Item Rules are applied to the Cart Item Total, and added to the Cart as Neutral Adjustment

solverat:
@dpfaffenbauer Btw, I was not able to test cart items action (`coreshop.cart_item_price_rule.action`). They never get applied (When using discount amount for example).

I don't know where the cart item rule processor should get triggered?

dpfaffenbauer:
@solverat They do work though for me. They are triggered as Cart Item Rule Action.

solverat:
On my debug marker, only `CoreShop\Component\Core\Cart\Rule\Action\DiscountAmountActionProcessor` gets dispatched, but never `CoreShop\Component\Core\CartItem\Rule\Action\DiscountAmountActionProcessor`.

dpfaffenbauer:
@solverat 
Screenshot 2024-03-05 at 10 40 26
Screenshot 2024-03-05 at 10 40 31
Screenshot 2024-03-05 at 10 42 11

solverat:
:see_no_evil: I'm so sorry, I'm very new to coreshop...

solverat (approved):

Issue #2568, opened 2024-02-27
Optionally dizactivate payment URL upon order fulfillment


TanaseTeofil:
| Q                | A
| ---------------- | -----
| Bug report?      | no
| Feature request? | yes
| BC Break report? | no
| RFC?             | no

Hello @dpfaffenbauer!

We have a proposal for a new feature that could be added to CoreShop and would probably improve security for many projects.
In CoreShop by Design you can access both the pay and thank-you pages for an unlimited period of time with a valid token.
This isn't particularly great because you can probably steal personal data by guessing the token. The more tokens there are in circulation, the easier it is to guess any of them. In practical terms, this isn't necessarily a problem with a long token.
Nevertheless, there is no reason why you should still see a Thank You page from an order after a year.
Would it be possible for you to solve this through time-limited or disposable tokens. This doesn't have to be the default but you could make it an option avaiable in CoreShop if you also believe it to be a good idea.

What are your thoughts on this?

dpfaffenbauer:
@TanaseTeofil Yes, I am down for that. Can you maybe describe it in more detail how you would achieve this?

TanaseTeofil:
@dpfaffenbauer I have talked a bit about this issue with @yariksheptykin and we reached the conclusion that we could maybe use JWT token generation since it provides a lot of features that we would need such as expiration date for the access tokens (configurable, but i think 15 minutes should be enough. Remains to be discussed). Initially we would just need the expiration date aspect of this but we could also use their encryption to further increase security for tokens on orders.

We can create a new token generator such as the UniqueTokenGenerator CoreShop already provides and add it as an alternative to the OrderFactory so that we can chose between the two through a config.

Give us a couple of days to whip up a Pull Request with a possible solution for this feature.
We will be collaborating with Iaroslav on this in the next days and we'll try to implement a solution for this.

dpfaffenbauer:
@TanaseTeofil lovely. Can't wait for your PR idea.

solverat:
-1 for JWT.

yariksheptykin:
@solverat could you write a couple of words why you don't think JWT is not a good idea?

solverat:
@yariksheptykin I think JWT is completely inflated for this task. 

- TTL: Why not simply check within the action points (pay, thank-you) if the order_date is < configurable_ttl and throw an exception, if so? 
- Token-Length: Changing the token length to a decent length could fix the "guessing" situation

TanaseTeofil:
[...]

PR #2574, opened 2024-03-05
Add time constrained order tokens

TanaseTeofil:
Closes https://github.com/coreshop/CoreShop/issues/2568


dpfaffenbauer:
@TanaseTeofil can you also please implement how you validate the token. I also think that JWT is a bit overkill. What if we use our Token Generator and just increase the length to 64 chars or so? this is for example how payum generates the tokens: https://github.com/Payum/Payum/blob/master/src/Payum/Core/Security/Util/Random.php#L21

TanaseTeofil:
> @TanaseTeofil can you also please implement how you validate the token. I also think that JWT is a bit overkill. What if we use our Token Generator and just increase the length to 64 chars or so? this is for example how payum generates the tokens: https://github.com/Payum/Payum/blob/master/src/Payum/Core/Security/Util/Random.php#L21

@dpfaffenbauer  That might be enough. I will discuss this with @yariksheptykin next week. Currently I am not available to work on this.

TanaseTeofil:
Hello @solverat !

We've decided to move forward with your suggestion on the issue #2568 
Can you provide a bit more details on how this configurable_ttl is configured and also how the tokenLength is decided?
Is token length something we give on generation as a parameter to the UniqueTokenGenerator only or is it configured somewhere else?
Any additional details you can provide on how you would implement this would be greatly appreciated.

TanaseTeofil:
> Hello @solverat !
> 
> We've decided to move forward with your suggestion on the issue #2568 Can you provide a bit more details on how this configurable_ttl is configured and also how the tokenLength is decided? Is token length something we give on generation as a parameter to the UniqueTokenGenerator only or is it configured somewhere else? Any additional details you can provide on how you would implement this would be greatly appreciated.

@dpfaffenbauer Can you help with this?

dpfaffenbauer:
@TanaseTeofil I would make it configurable with a symfony config and inject it with DI.

PR #2603, opened 2024-04-09
Index error if product does not have a store

hethehe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   none

If there is a product without a store, the coreshop:index fail with:
`implode(): Argument #1 ($array) must be of type array, string given`


dpfaffenbauer (approved):

dpfaffenbauer:
thanks @hethehe

PR #2609, opened 2024-04-19
make password reset notification store aware

hethehe:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | none

There is an error if you use the password reset notification with store based emails.
`Message with ID "62164" of Class "CoreShop\Component\Notification\Messenger\NotificationMessage" failed. Redelivery tried at "2024-04-19 11:47:40". Error was: "StoreMailActionProcessor: Store is not set."`
=> That's because the Customer is not an instance of StoreAwareInterface.

With this PR the store_id is added to the rules parameters.


dpfaffenbauer:
something is breaking the behat tests, I am pretty certain its not your change. I will check next week

dpfaffenbauer:
reason this fails is this: https://github.com/coreshop/CoreShop/pull/2612. Your PR is fine, thanks @hethehe

Issue #2621, opened 2024-04-26
Removing Unit Definition leads to corrupt Product Unit Definition State


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Tested on Coreshop 3.x, 4.x

Reproducible on demo:
- open product with id `105`
- go to "price & units" and remove additional unit "palette"

![image](https://github.com/coreshop/CoreShop/assets/700119/d360c04f-0fa4-4a44-894e-c0d5ceb229f2)
![image](https://github.com/coreshop/CoreShop/assets/700119/0b5aa48f-6849-4423-8763-69560fc9c352)
![image](https://github.com/coreshop/CoreShop/assets/700119/884b0ecc-1526-41fd-8dd7-fb41d3309cfd)

- save product and reopen it
- go to "price & units":

![image](https://github.com/coreshop/CoreShop/assets/700119/7f883144-7505-4098-8239-0c9d573f2062)
![image](https://github.com/coreshop/CoreShop/assets/700119/eda9f15c-282c-40e5-a4a8-798c34574e70)
![image](https://github.com/coreshop/CoreShop/assets/700119/b11b0349-aa76-427b-9b26-f1ec4885e8d0)

Bug:
- All definitions in `coreshop_product_unit_definition` are gone
- The main entry in `coreshop_product_unit_definitions` stays orphaned which leads to an "duplicate entry" if you try to re-add unit definitions again

solverat:
It's definitely something within the `EntityMerger`.

Also some related/unrelated logs:

User Deprecated: While adding an entity of class Proxies\__CG__\CoreShop\Component\Product\Model\ProductUnitDefinitions with an ID hash of "7" to the identity map, another object of class CoreShop\Component\Product\Model\ProductUnitDefinitions was already present for the same ID. This will trigger an exception in ORM 3.0.  IDs should uniquely map to entity object instances. This problem may occur if:  - you use application-provided IDs and reuse ID values; - database-provided IDs are reassigned after truncating the database without clearing the EntityManager; - you might have been using EntityManager#getReference() to create a reference for a nonexistent ID that was subsequently (by the RDBMS) assigned to another entity.

PR #2622, opened 2024-04-27
[ResourceBundle] Fix EntityMerger and cascade-merging

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2621

@solverat Please check if this works.

We need to refactor these things completely. It is so hard to debug and understand. I remember why we have the EntityMerger, but I am certain there are easier ways. 

Main reason for the Entity Merger is:

 - In preparation of the unsaved data-object, we don't use the EntityManager to load the data.
 - Reason is: If you do changes and decide to not persist them, EntityManager holds the wrong state. Any flush on it, will cause the "temporary" changes to be flushed to the DB.
 - This is one big disadvantages of Entity Management. Every state change that gets persisted, gets flushed.
 - Since we don't know if a change gets persisted or not, we can't even use `refresh` to get the "original" data.
 - So the merger gets the data from the actual Entity Manager, and merges them with our changes.
 

One workaround might be to use DTOs. So we don't actually work with real entities in combination with Pimcore DataObjects, but with DTOs that get converted to Entity on persisting. Not sure about it though.

### DIFF
--- a/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
@@ -330,6 +330,11 @@ class EntityMerger
                 $id = $relatedEntityClass->getIdentifierValues($relatedEntities);
                 $uwEntity = $this->em->getUnitOfWork()->tryGetById($id, $relatedEntityClass->getName());
 
+                //Entity might not be loaded and managed yet, try to load it
+                if (!$uwEntity) {
+                    $uwEntity = $this->em->find($relatedEntityClass->getName(), $id);
+                }
+
                 if ($uwEntity) {
                     $class->reflFields[$assoc['fieldName']]->setValue($entity, $uwEntity);
                 }
--- a/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
@@ -330,7 +330,9 @@ class EntityMerger
                 $id = $relatedEntityClass->getIdentifierValues($relatedEntities);
                 $uwEntity = $this->em->getUnitOfWork()->tryGetById($id, $relatedEntityClass->getName());
 
-                $class->reflFields[$assoc['fieldName']]->setValue($entity, $uwEntity);
+                if ($uwEntity) {
+                    $class->reflFields[$assoc['fieldName']]->setValue($entity, $uwEntity);
+                }
             }
         }
 
--- a/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
@@ -313,6 +313,10 @@ class EntityMerger
         foreach ($noMergeAssociationMappings as $assoc) {
             $relatedEntities = $class->reflFields[$assoc['fieldName']]->getValue($entity);
 
+            if (!$relatedEntities) {
+                continue;
+            }
+
             if ($relatedEntities instanceof Collection) {
                 //Reset Collection
                 $pColl = new PersistentCollection($this->em, $assoc['targetEntity'], new ArrayCollection());
@@ -320,7 +324,7 @@ class EntityMerger
                 $pColl->setInitialized(false);
 
                 $class->reflFields[$assoc['fieldName']]->setValue($entity, $pColl);
-            } elseif ($relatedEntities !== null) {
+            } else {
                 //Reset "tmp" entity with managed entity
                 $relatedEntityClass = $this->em->getClassMetadata($assoc['targetEntity']);
                 $id = $relatedEntityClass->getIdentifierValues($relatedEntities);
--- a/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
+++ b/src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
@@ -300,7 +300,39 @@ class EntityMerger
                 return $assoc['isCascadeMerge'];
             }
         );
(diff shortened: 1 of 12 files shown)


dpfaffenbauer:
It also seems to break other stuff...

dpfaffenbauer:
@solverat this works, but might break versions. Can you check please?

solverat:
@dpfaffenbauer Thanks for your investigation! It is simply brutal. Getting rid of the EntityMerger would be a good thing - especially for DX! 

I just added your changes and removing / updating / adding units works flawless.

If I'm remembering it right, the EntityMerger exists because of Versions to handle drafts. With these changes, creating / updating / publishing versions are not working anymore (Which is quite obvious).

***

So I tried to go further without using the Entity Manager. I did two things:

- get entity from tmp manager if it's a version task (I don't like this tmp manager, but I guess there is no other solution as you explained above)
- using unmarshalVersion like a new form submission

First, if we're storing a version, do not use the default entity manager in `getDataFromEditmode`:

```php
public function getDataFromEditmode(mixed $data, Concrete $object = null, array $params = []): mixed
{
    if (!is_array($data)) {
        return null;
    }

    $errors = [];
    
    // example code here!
    $onlyVersion = \Pimcore::getContainer()->get('request_stack')->getMainRequest()->query->get('task') === 'version';

    if ($onlyVersion === false) {
        $unitDefinitionsEntity = $this->getProductUnitDefinitionsRepository()->findOneForProduct($object);
    } else {
        $tempEntityManager = $this->createTempEntityManager($this->getEntityManager());
        $tempStoreValuesRepository = $this->getProductUnitDefinitionsRepositoryFactory()->createNewRepository($tempEntityManager);
        $unitDefinitionsEntity = $tempStoreValuesRepository->findOneForProduct($object);
    }

    // [...]
```

Second, we need to adjust the unmarshal method. PLEASE! Do not judge the example code here, I just modified the serialized data to simulate the form data:

```php
public function unmarshalVersion(Concrete $object, mixed $data): mixed
    {
        if (!is_array($data)) {
            return null;
        }

        $unitDefinitionsEntity = $this->getProductUnitDefinitionsRepository()->findOneForProduct($object);
        $form = $this->getFormFactory()->createNamed('', ProductUnitDefinitionsType::class, $unitDefinitionsEntity);

        // @todo: use marshalVersion to store data in a right manner so we don't have to do this mess:

        $data['product'] = $object->getId();

        $additional = [];
        foreach($data['unitDefinitions'] as $index => $unitDefinition) {
            if($unitDefinition['id'] === $data['defaultUnitDefinition']['id']) {
                continue;
            }

            $unitDefinition['unit'] = $unitDefinition['unit']['id'];
            
            $additional[$index] = $unitDefinition;
        }

        $data['additionalUnitDefinitions'] = $additional;
        $data['defaultUnitDefinition'] = [
            'unit' => $data['defaultUnitDefinition']['unit']['id'],
            'precision' => $data['defaultUnitDefinition']['precision'],
            'conversionRate' => $data['defaultUnitDefinition']['conversionRate'],
        ];

        unset($data['unitDefinitions']);

        $form->submit($data);

        return $form->getData();
    }
```

The EntityManager is required by following data types and if this is a suitable way, we should also take care about them:

- ProductSpecificPriceRules
- ProductQuantityPriceRules
- StoreValues
- ProductUnitDefinitions

dpfaffenbauer:
@solverat Main issue is how Pimcore persists and in what order. We do run the flush multiple times due to how Pimcore works. We run it for StoreValues, UnitDefinitions and Price Rules.

In this scenario, the unit definitions are the first. They actually get deleted, but then the StoreValues one comes and runs and re-adds it. But not completely right. 

There is 2 fixes in here, one for only merge cascades that are actually configured to fix the issue with the ordering and old "entity" state. I had the issue that the Unit Definition never got deleted, the Unit Definitions Persister deleted it, the StoreValues added it back again. Since that one has a relation to it via Unit Definition Pricing. If we don't cascade merge that relation, we don't get the issue.

It is all a bit buggy and tricky...

src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
```diff
@@ -92,6 +88,8 @@ private function doMerge($entity, array &$visited): void
                     $this->cascadeMerge($entity, $visited);
                     $this->em->getUnitOfWork()->persist($entity);
                 } else {
+                    $managedCopy = $this->loadEntity($managedCopy);
```
dpfaffenbauer, line 91: @solverat here was the issue with the lost/broken Unit-Definitions

src/CoreShop/Bundle/ResourceBundle/Doctrine/ORM/EntityMerger.php
```diff
@@ -278,7 +294,14 @@ private function cascadeMerge($entity, array &$visited): void
     {
         $class = $this->em->getClassMetadata($entity::class);
 
-        foreach ($class->associationMappings as $assoc) {
+        $associationMappings = array_filter(
```
dpfaffenbauer, line 297: @solverat this and the mapping files was the reason for the re-added unit-definition

solverat:
There seems to be an issue if you're trying to remove an additional unit.

- add two units to product
- save
- remove one unit
- save

=> checkout db:
- unit is still there (has been readded, check ID)

=> next:
- change quantity of remaining unit
- save
- the removed unit is finally gone

dpfaffenbauer:
@solverat this is killing me... Maybe, only maybe, I got it this time.

solverat:
Improved. Next (Versions):

- Change unit
- Create draft
- Publish draft

Exception: Warning: Attempt to read property "reflFields" on string

solverat:
Also: QPR Ranges can't be added anymore (add a qpr with a range and unit and save).

dpfaffenbauer:
@solverat please try again

PR #2625, opened 2024-05-07
Translatable labels for payment-provider-rules

nehlsen:
Added translatable labels for payment-provider-rules

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #...   

BEFORE
coreshop-add-localized-label-to-payment-provider-rules-before

AFTER
coreshop-add-localized-label-to-payment-provider-rules-after


src/CoreShop/Bundle/CoreBundle/Migrations/Version20230829163505.php
```diff
@@ -0,0 +1,42 @@
...
+final class Version20230829163505 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('
```
dpfaffenbauer, line 37: can wee add checks if the table already exists?

dpfaffenbauer (changes requested):
lovely work, just the migration and then we are good to go.

Thank you very much @nehlsen

src/CoreShop/Bundle/CoreBundle/Migrations/Version20230829163505.php
nehlsen: Thank You!
I have added a check. does it suffice?

dpfaffenbauer: Can you rather do something like

`if (!$schema->hasTable('coreshop_payment_provider_rule_translation') {return}`

That way the add constraint can also not fail anything.

nehlsen: sure thing :)

dpfaffenbauer (approved):

dpfaffenbauer:
thanks a lot @nehlsen

Issue #2615, opened 2024-04-21
Payment Manuall Confirmation doesn't lead to Order Confirmation


dpfaffenbauer:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? |no
| RFC?             |no

If you get into the situation that the payment provider never notified you about the payment state, and therefore your payment is stuck in "pending". If you then manually "confirm" the order using the backend, the order never gets confirmed.

What if we bind the order-state to the payment state? or at least, check on the payment transition for the order state and trigger the confirm? We currently only trigger the confirm state on payment with payum.

@solverat what do you think?

PR #2628, opened 2024-05-07
[Payment] when payment is authorized or paid, order get's confirmed

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2615

Current Issue is that:
 1. `UpdateOrderStateExtension` never updates to order state cause the payment state got changed before
 2. if you manually executed the `pay` transition, order is stuck in `new` and does not get confirmed

This PR automatically `confirms` the order when a single payment is partially paid, fully paid, partially authorized or fully authorized.

@solverat I would like to get your opinion on this.


dpfaffenbauer:
@solverat We would like to get your Feedback on this.

solverat:
I'm afraid of this.

solverat:
Funny! We just experienced this behavior in one of our new shops (not in manually mode but by payum / client workflow).

image

I'm unable to reproduce it, though...

Not sure, if this is true: 

> UpdateOrderStateExtension never updates to order state cause the payment state got changed before

The only difference now is, that the callback happens after a transition instead of the payum extension. If we're dealing with a sort of race condition here, this PR could be an improvement, since the order transition will be dispatched at the end of the payum cycle.

dpfaffenbauer:
@solverat so not afraid anymore? :D

solverat:
I guess not 😀

dpfaffenbauer:
@solverat the main issue that this tries to solve is, when a user exists the payment workflow just before we come back to the payum capture/authorize route, and there is no async implemented, you never know the status of the payment. if you then manually confirm it in the backend using the payment workflow, the order never gets confirmed. this fixes that. 

We actually could reproduce it. We looged at the access logs of these orders (based on timestamps) and saw they never returned from the PSP to the shop.

Issue #2642, opened 2024-05-30
The service "CoreShop\Bundle\MessengerBundle\Controller\ListMessagesController" has a dependency on a non-existent service "Pimcore\Security\User\TokenStorageUserResolver".


solverat:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

Release [3.2.8](https://github.com/coreshop/CoreShop/releases/tag/3.2.8) is broken. Don't know why tests have passed here https://github.com/coreshop/CoreShop/commit/1a0e629e8053d757eb01f63c9abceb68fb38539d since this will crash the whole application?

dpfaffenbauer:
@solverat I keep this open since until the min requirement of 10.6 is merged

PR #2643, opened 2024-05-30
fix token resolver service

solverat:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #2642

### DIFF
--- a/src/CoreShop/Bundle/MessengerBundle/Resources/config/services.yml
+++ b/src/CoreShop/Bundle/MessengerBundle/Resources/config/services.yml
@@ -57,11 +57,11 @@ services:
       - { name: controller.service_arguments }
     calls:
       - [setContainer, ['@service_container']]
-      - [setTokenResolver, ['@Pimcore\Security\User\TokenStorageUserResolver']]
+      - [setTokenResolver, ['@Pimcore\Bundle\AdminBundle\Security\User\TokenStorageUserResolver']]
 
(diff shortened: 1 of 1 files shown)


dpfaffenbauer:
I'll further review this so the tests break when this happens in the future

dpfaffenbauer:
@solverat CoreShop 3.2 is Pimcore 10 only, that service is available there: https://github.com/pimcore/pimcore/blob/v10.6.9/lib/Security/User/TokenStorageUserResolver.php

It also exists in the Pimcore 11.2 branch for CoreShop 4.x: https://github.com/pimcore/pimcore/blob/11.2/lib/Security/User/TokenStorageUserResolver.php

dpfaffenbauer:
@solverat could it be that you are running pimcore 10.5?

solverat:
god damnit, yes. We're on 10.5.25 on some projects. since this is a really stable version, we skipped the "deprecation release".

dpfaffenbauer:
@solverat see: https://github.com/coreshop/CoreShop/pull/2644

we are only testing against 10.6 and therefore can only guarantee to be compatible with that. therefore we upgrade the dependencies to 10.6 only.

dpfaffenbauer:
> god damnit, yes. We're on 10.5.25 on some projects. since this is a really stable version, we skipped the "deprecation release".

10.6.9 is also stable ;)

Issue #2676, opened 2024-08-07
Price rule actions affect all product unit prices


EinShoppo:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | n

**Steps to reproduce:**
Create a product with different product units, each product unit has a different price.
![image](https://github.com/user-attachments/assets/46ab94a2-02d0-49a4-ada7-26a656f9d65e)

Add a product specific price rule to the product which sets the price to 100 EUR. 
![image](https://github.com/user-attachments/assets/9ff1e6cb-fcd7-45f7-8ccd-49e4920e9e15)

**Actual**
The price of _all_ product units is changed to the price from the price rule (=100 EUR).

**Expected:**
Only the price of the default product unit should be changed or – even better – there is a way to set the price for certain product units. Or did I miss something?
May be the problem could be solved directly within the action by adding a product unit selection. The problem also exists with all kind of actions (discount price, discount amount, discount percent, not discountable) and probably also with other types of price rules, such as the general product price rules.
![image](https://github.com/user-attachments/assets/fff3fc0a-b7b9-4e50-a472-9586945c41be)

dpfaffenbauer:
@EinShoppo we are aware of that, we never added pricing rule conditions for units. One way is, as you already mentioned, do have a unit selector, but what would that selector do for products without a unit? Percentage discounts might still be valid for products with units.

EinShoppo:
@dpfaffenbauer Okay, I see. I think the selector should contain a wildcard value (null) so that you can apply an action to all units. In this way, products without units could also be handled. I personally believe that percentage discounts should also be limitable on certain units, because you might want to discount the unit "carton" by 10%, for example, but not the unit "piece".
A multi-select for selecting the units would probably be ideal, then you can select several units with one action if required.

dpfaffenbauer:
yes. Problem might be that the rule Action itself doesn't know anything about the unit. That is something we need to check and solve first.

dpfaffenbauer:
@EinShoppo we fixed it for now to disable Price and DiscountPrice for units: https://github.com/coreshop/CoreShop/pull/2678

dpfaffenbauer:
@EinShoppo The bug fix is merged, I will keep this ticket open as a feature request for the unit selection

PR #2678, opened 2024-08-08
[Product] don't allow Price and DiscountPrice Rule for unit prices

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | no
| Deprecations? |no
| Fixed tickets | #2676


solverat:
@dpfaffenbauer Since this commit, the price rules don't get applied when adding a product with a simple price rule to the cart. This is a quite huge issue for us at this very moment, since we have to remove those lines from all SLA projects - where some of them had some serious discount campaigns up and running.

PR #2685, opened 2024-08-14
ReplyToSymfonyResponseConverter is moved into `PayumBundle` from `Payum\Core\Bridge`

aashan10:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | 

This file is moved into the bundle with a new namespace which needed fixing:

https://github.com/Payum/PayumBundle/blob/master/ReplyToSymfonyResponseConverter.php
https://github.com/Payum/PayumBundle/commit/0180a1f888311172569d34c31997b32b803dae12


dpfaffenbauer:
@aashan10 Payum did not yet create another release with this change, therefore we cannot change this yet in our code.

Issue #2699, opened 2024-09-05
Order Payment Total - wrong formatting when using different decimal factor


MatejaPodravac:
| Q                | A
| ---------------- | -----
| Bug report?      | yes
| Feature request? | no
| BC Break report? | no
| RFC?             | no

On the order overview page in Pimcore backend in the summary section order payment total is not displayed correctly. For example, total is 44€ and payment total stored is 4400. 

Configuration for fraction digits:
```
core_shop_currency:
    money_decimal_precision: 2
    money_decimal_factor: 10000
```

coreshop.util.format.currency relies completely on the defined decimal factor which results in payment total not being displayed correctly. Instead of defined decimal factor, it should rely on the factor passed from OrderController `getSummary` for every summary item. Only if it's not passed use defined fraction digits.

![image](https://github.com/user-attachments/assets/feff9b77-e561-4efa-9d6d-aab26e0b8036)
![image](https://github.com/user-attachments/assets/edd15b71-d963-4f5b-b601-c4882bbf8552)

MatejaPodravac:
@dpfaffenbauer do you plan to include this fix in the 4.0.10 release?

dpfaffenbauer:
@MatejaPodravac it is, every change in 3.2.x gets merged up to 4.0.x and next

PR #2700, opened 2024-09-05
[OrderBundle] fix precison/factor for Payment Total

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2699

### DIFF
--- a/src/CoreShop/Bundle/OrderBundle/Resources/public/pimcore/js/order/detail/blocks/detail.js
+++ b/src/CoreShop/Bundle/OrderBundle/Resources/public/pimcore/js/order/detail/blocks/detail.js
@@ -268,7 +268,17 @@ coreshop.order.order.detail.blocks.detail = Class.create(coreshop.order.order.de
                     dataIndex: 'value',
                     width: 150,
                     align: 'right',
-                    renderer: function (value) {
+                    renderer: function (value, metaData, record) {
+                        var data = record.data;
+                        if (data.hasOwnProperty('precision') && data.hasOwnProperty('factor')) {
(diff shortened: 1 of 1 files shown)


MatejaPodravac:
@dpfaffenbauer shouldn't it be if it doesn't have own property to use global, otherwise use property from data?

src/CoreShop/Bundle/OrderBundle/Resources/public/pimcore/js/order/detail/blocks/detail.js
```diff
@@ -268,7 +268,17 @@ coreshop.order.order.detail.blocks.detail = Class.create(coreshop.order.order.de
                     dataIndex: 'value',
                     width: 150,
                     align: 'right',
-                    renderer: function (value) {
+                    renderer: function (value, metaData, record) {
+                        var data = record.data;
+                        if (data.hasOwnProperty('precision') && data.hasOwnProperty('factor')) {
```
MatejaPodravac, line 273: @dpfaffenbauer this doesn't look right, if it does have own property use it, otherwise use global

dpfaffenbauer: yes

dpfaffenbauer: fixed

MatejaPodravac (approved):
thanks for the fix

PR #2728, opened 2024-10-16
Revert "[Twig] twig >= 3.14 is compatible again"

dpfaffenbauer:
Reverts coreshop/CoreShop#2719

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -139,7 +139,7 @@
   },
   "conflict": {
     "jms/serializer-bundle": "4.1.0",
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
Pimcore actually did not yet allow it, so we revert it too.

fashxp:
https://github.com/pimcore/pimcore/pull/17593 ?

dpfaffenbauer:
WTF, I just checked the composer.json and didn't see it :D. 

Thanks for pointing out, I will revert the revert then.

PR #2781, opened 2025-01-16
Do not load complete product again from database on pre_update

BlackbitDevs:
When a product gets saved, currently during `pimcore.dataobject.preUpdate` event the whole object gets loaded again from database - only to find out if the `published` state got changed. This is quite expensive from performance perspective. This PR changes this with a direct SQL query. I am quite sure that you will not accept it because of the SQL inside an event listener but maybe we can find a clean solution...

### DIFF
--- a/src/CoreShop/Bundle/CoreBundle/EventListener/ProductAvailabilityEventListener.php
+++ b/src/CoreShop/Bundle/CoreBundle/EventListener/ProductAvailabilityEventListener.php
@@ -50,27 +50,16 @@ final class ProductAvailabilityEventListener
             return;
         }
 
-        if (in_array($object->getId(), $this->productIdsToCheck, true)) {
-            return;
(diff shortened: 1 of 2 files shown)


dpfaffenbauer:
@BlackbitDevs the SQL inside the listener is fine, but can you Dependency Inject the SQL Connection and not use \Pimcore\DB ?

dpfaffenbauer:
ping @BlackbitDevs

dpfaffenbauer:
ping @BlackbitDevs

BlackbitDevs:
Sorry for the late reaction. Still trying to catch up my todo list after christmas vacation ;-)

dpfaffenbauer (approved):

dpfaffenbauer:
@BlackbitDevs thanks a lot :)

PR #2808, opened 2025-02-27
Update .phpda

aziel18:
| Q             | A
| ------------- | ---
| Bug fix?      | yes/no
| New feature?  | yes/no
| BC breaks?    | no
| Deprecations? | yes/no
| Fixed tickets | #...


dpfaffenbauer:
@aziel18 Thanks for the PR, we actually haven't used this since forever, so we might better delete it. Whats the point of updating it from your point of view?

PR #2821, opened 2025-04-14
[Core] fix for gift digital product (revert bc break)

dpfaffenbauer:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  |no
| BC breaks?    | no
| Deprecations? |no

https://github.com/coreshop/CoreShop/pull/2820 accidentally introduced a bc break - this fixes it

### DIFF
--- a/src/CoreShop/Component/Order/Model/PurchasableInterface.php
+++ b/src/CoreShop/Component/Order/Model/PurchasableInterface.php
@@ -29,6 +29,4 @@ interface PurchasableInterface extends ResourceInterface
     public function getWholesaleBuyingPrice(): ?Money;
 
     public function getTaxRule(): ?TaxRuleGroupInterface;
-
-    public function getDigitalProduct(): ?bool;
 }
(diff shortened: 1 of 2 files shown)


src/CoreShop/Component/Order/Model/PurchasableInterface.php
```diff
@@ -29,6 +29,4 @@ public function getName($language = null): ?string;
     public function getWholesaleBuyingPrice(): ?Money;
 
     public function getTaxRule(): ?TaxRuleGroupInterface;
-
-    public function getDigitalProduct(): ?bool;
```
jdreesen, line 33: Symfony adds new methods via `@method bool|null getDigitalProduct()` PHPDoc on the interface before adding it for real in the next major version, so people can already see that it exists.

dpfaffenbauer: @jdreesen thanks, but that method should not be there at all.

PR #2840, opened 2025-05-07
[IndexBundle] Add OpenSearch Index Worker

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

This PR adds support to CoreShop Index to use OpenSearch as its index engine. It integrates with the [Pimcore OpenSearch Client Bundle](https://github.com/pimcore/opensearch-client) to configure one or multiple OpenSearch clients.

### Index settings
Since OpenSearch's index settings are complex and have many different options, I've included the possibility to hook into the creation/update process via an event listener. This should make it very easy to extend the default configuration, which is passed on to the OpenSearch API.

To start, I've added the following config options:
- Number of Shards
- Number of Replicas

### Index Mapping
The mapping is following best practices with "explicit mapping" of all fields. The [field types](https://docs.opensearch.org/docs/2.19/field-types/#dynamic-mapping-types) as well as [language analyzers](https://docs.opensearch.org/docs/2.19/analyzers/language-analyzers/index/) are added automatically.


aarongerig:
I have read the CLA Document and I hereby sign the CLA

aarongerig:
recheck

dpfaffenbauer (approved):

PR #2854, opened 2025-06-19
[Messenger] dispatch `MessageDetailsEvent` to allow customization of message details generation

jdreesen:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no

Currently, the “Details” modal only contains a `print_r` of the message. Sometimes, however, it is desirable to present the content better. For this purpose, a `MessageDetailsEvent` has been introduced, which can be used to manipulate the displayed data. 

Here is an example for a message that contains XML:

```php
#[AsEventListener('coreshop.messenger.message_details')]
class MyMessageDetailsListener
{
    public function __invoke(MessageDetailsEvent $event): void
    {
        if ('my_receiver' === $event->getReceiverName()) {
            $xml = $event->getEnvelope()->getMessage()->xml;

            $event->getMessageDetails()->setSerialized('<pre>'.htmlentities($xml).'</pre>');
        }
    }
}
```

PR #2876, opened 2025-07-08
[Messenger] Fix `MessageRepository` service registration

jdreesen:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #2854

Fixes the service configuration with the new parameter that got added in #2854.

It would be great if we could get a bug fix release for `3.x` soon, as the `MessengerBundle` is currently not really usable (you get an exception when opening it in the Pimcore backend). 

Note: this was already fixed for `4.x` in 918f626a2fcaa4a7738e5dc1093ba88fe8fd9f8e, so it doesn't need a new release.


dpfaffenbauer (approved):

dpfaffenbauer:
@jdreesen done and released 3.2.20

PR #2880, opened 2025-07-11
[Index] CoreShop Index Command with Opensearch

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

OpenSearch does not allow to delete documents that do not exist in the index. Therefore we need to check whether the document exists before trying to delete it.

### DIFF
--- a/src/CoreShop/Bundle/IndexBundle/Worker/OpenSearchWorker.php
+++ b/src/CoreShop/Bundle/IndexBundle/Worker/OpenSearchWorker.php
@@ -198,6 +198,17 @@ class OpenSearchWorker extends AbstractWorker implements OpenSearchWorkerInterfa
 
         // Delete the document from the index if it is not indexable
         if (!$object->getIndexable($index)) {
+            $documentExists = $client
+                ->exists([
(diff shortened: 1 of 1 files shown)


src/CoreShop/Bundle/IndexBundle/Worker/OpenSearchWorker.php
```diff
@@ -198,6 +198,17 @@ public function updateIndex(IndexInterface $index, IndexableInterface $object):
 
         // Delete the document from the index if it is not indexable
         if (!$object->getIndexable($index)) {
+            $documentExists = $client
+                ->exists([
```
dpfaffenbauer: why not just try catch to avoid the second round-trip?

aarongerig: @dpfaffenbauer You're totally right – I changed it.

dpfaffenbauer (approved):

dpfaffenbauer:
❤️

PR #3028, opened 2026-05-01
[IndexBundle] OpenSearchWorker: Refactor delete methods for index operations

aarongerig:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no

We experience a lot of failed `IndexDeleteMessages` with the following error:
```log
{"_index":"coreshop_index_os_INDEX_NAME","_id":"1234","_version":1,"result":"not_found","_shards":{"total":1,"successful":1,"failed":0},"_seq_no":1234,"_primary_term":8}
```

Whenever the OpenSearch worker tries to delete a non-existent document in the index, the operation fails as well as the message handling process. This clogs the failed transport of `coreshop_index` and therefore we'd like to introduce this change.

With this PR, we first check if the ID exists in the index before attempting to remove that index document. If it doesn't exist, we abort the deletion process and all is good. 🙂

### DIFF
--- a/src/CoreShop/Bundle/IndexBundle/Worker/OpenSearchWorker.php
+++ b/src/CoreShop/Bundle/IndexBundle/Worker/OpenSearchWorker.php
@@ -162,12 +162,17 @@ class OpenSearchWorker extends AbstractWorker implements OpenSearchWorkerInterfa
      */
     public function deleteFromIndexById(IndexInterface $index, int $id): void
     {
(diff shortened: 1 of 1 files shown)


dpfaffenbauer (approved):

dpfaffenbauer:
thanks, release next week or the week after

PR #3124, opened 2026-07-27
[Studio] Render CoreShop trees with Pimcore's TreeElement

dpfaffenbauer:
## What

CoreShop's trees each rendered a hand-rolled antd `Tree` with their own `.ant-tree-*` overrides, so they looked nothing like Pimcore's trees — no icon slot, doubled row height, different hover/selected colors and the default antd switcher triangle.

They now use `TreeElement` from `@pimcore/studio-ui-bundle/components`, the same component Pimcore uses for its Classes, Thumbnails and Users trees:

| Component | Change |
|---|---|
| `ResourceBundle` `EntityList` | `TreeElement`; icons via `icon`, delete via node `actions`, counts in the title string |
| `IndexBundle` `ClassDefinitionTree` | `TreeElement`, `Draggable` via `titleRender`, new "Add Field" context action |
| `IndexBundle` `SelectedFieldsTree` | `TreeElement`, edit/delete as node actions instead of a custom antd `Dropdown` |
| `IndexBundle` `field-type-icons.ts` | new — the field-type→icon map was duplicated verbatim in both index trees |
| `VariantBundle` `VariantGeneratorModal` | checkable tree switched over, folder/leaf icons added |

Titles are plain strings the way Pimcore does it (`Notification Rules (7)`, `name (input)`), so no wrapper elements are needed for decoration. `titleRender` is only used where a wrapper is unavoidable — drag-and-drop and double-click.

## Module federation

`antd`, `antd-style`, `@ant-design/colors`, `@ant-design/icons`, `classnames` and `dayjs` are now shared as singletons. Without that, every CoreShop remote instantiated its own antd and rendered outside Pimcore's `ConfigProvider`/antd-style theme, falling back to antd's default tokens — which is why CoreShop controls were blue instead of the Pimcore palette.

## Notes

- Pimcore's `Draggable`/`Droppable` render a block-level `div` inside `.ant-tree-title`. Nodes wrapped in one carry `ant-tree-node--has-drag-and-drop` plus a single `height: 24px` rule on the title button — the same fix Pimcore applies to its own `.hotspot-droppable`.
- The default `leafIcon` was `widget-default`, which does not exist in Pimcore's icon library (only `widget.inline.svg`), so leaves rendered without an icon.
- Hardcoded English strings in the index trees now use translation keys (`studio.en.yaml` / `studio.de.yaml`).
- `npm run clean:builds` (`studio-clean.ts`) drops generated Studio build output locally; `--dry` lists what it would remove.

## Verification

- `tsc --noEmit` clean for Resource, Index and Variant bundles
- full `npm run build` green
- docs updated in `docs/03_Development/14_Studio/02_Base_Infrastructure/01_ResourceBundle.md`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dpfaffenbauer:
Superseded — reopened from a branch on coreshop/CoreShop instead of a fork.

PR #3219, opened 2026-09-04
[Telemetry] Move telemetry into its own bundle so it works without CoreShopCoreBundle

### LINKED ISSUE #3218: [Telemetry] Move telemetry into its own bundle so it works without CoreShopCoreBundle
## Problem

The telemetry ping added in #3213 lives in `CoreShopCoreBundle` (`CoreShop\Bundle\CoreBundle\Telemetry\*`, contracts in `CoreShop\Component\Core\Telemetry\*`). Several CoreShop bundles are used on plain Pimcore installations without `CoreShopCoreBundle`: `coreshop/document-route-bundle`, `coreshop/inbound-email-rules-bundle`, `coreshop/batch-messenger-bundle` and `coreshop/index-geo-bundle`.

On such an installation the enterprise subscription bundle cannot be loaded: it depends on `TelemetryPingerInterface`, which is only registered by the CoreBundle extension. The result is a container compile error, and those installations never send a ping or receive a license status.

## Wanted

- Telemetry contracts and implementation in their own bundle (`coreshop/telemetry-bundle`) with no dependency on CoreBundle or ResourceBundle.
- `CoreShopCoreBundle` registers the telemetry bundle as a dependent bundle, so full installations keep working unchanged.
- Configuration moves from `core_shop_core.telemetry` to `core_shop_telemetry`; the old node has not been released yet, so no compatibility layer is needed.
- Both lines: 2026.x and 5.1.

dpfaffenbauer:
Closes #3218

## New bundle `coreshop/telemetry-bundle`

`src/CoreShop/Bundle/TelemetryBundle` (`CoreShop\Bundle\TelemetryBundle\CoreShopTelemetryBundle`, extends `AbstractPimcoreBundle`, no ResourceBundle/CoreBundle dependency). The four contracts moved from `CoreShop\Component\Core\Telemetry\*` to `CoreShop\Bundle\TelemetryBundle\Contract\*` with unchanged signatures; implementations live in `Identity\InstanceIdentifierProvider`, `Storage\SettingsStoreTelemetryResultStorage`, `Provider\{System,Hosts,Bundles}DataProvider`, `Pinger\TelemetryPinger`, `Maintenance\TelemetryTask`, `Command\TelemetryPingCommand`. `Version\CoreShopVersion` resolves the CoreShop version through `class_exists` on the CoreBundle `Version` class and falls back to `Composer\InstalledVersions` (`coreshop/core-shop`, then `coreshop/telemetry-bundle`), so the bundle works on installations that only run single CoreShop bundles. Package requires `pimcore/pimcore`, `symfony/http-client`, `symfony/uid`.

## Configuration and wiring

Root key is now `core_shop_telemetry { enabled, endpoint, timeout }` (`CoreShopTelemetryExtension`, parameters `coreshop.telemetry.*`, `registerForAutoconfiguration` for the provider tag, `TaggedIteratorArgument` wiring as before, `services_test.yml` with the `MockHttpClient` override loaded when the environment contains `test`). `CoreShopCoreBundle` drops the telemetry node, parameters, service import and test override and registers `CoreShopTelemetryBundle` as a dependent bundle (priority 1300). `core_shop_core.telemetry` was never part of a release, so there is no compatibility layer. Monorepo plumbing: `replace` entry and `.gitsplit.yml` entry for `coreshop/telemetry-bundle`.

## Payload addition

`InstanceIdentifierProviderInterface::getRawPimcoreIdentifier()` returns `PIMCORE_INSTANCE_IDENTIFIER` in clear text; the pinger sends it as `pimcoreInstanceIdentifier` so the portal can match installations with Pimcore's product registration. The generated UUID fallback is never exposed there; `id` (sha256) and `pimcoreInstanceId` (hmac) are unchanged. Docs page and changelog updated.

## Not in this PR

The 5.1 backport is a separate PR against `5.1`.

## Test

- `vendor/bin/ecs check` on `src/CoreShop/Bundle/TelemetryBundle` and the CoreBundle changes: no errors
- `vendor/bin/phpstan analyse -c phpstan-package.neon --level 3` on the same paths: no errors
- `vendor/bin/psalm` on the same paths: no errors
- Not verified: kernel boot / `coreshop:telemetry:ping --dump` in the worktree, because the locally installed vendor (`pimcore/studio-ui-bundle` 2026.1) is older than what `2026.x` now requires and the container fails on `BuildArchiveExtractionTrait` before reaching the telemetry services. The moved code is byte-identical apart from namespaces and the `pimcoreInstanceIdentifier` field; the previous wiring was verified in #3214.


dpfaffenbauer:
Superseded: the TelemetryBundle was merged on 5.1 (#3220) and reaches 2026.x through the regular upmerge instead of this separate PR.

